PR #10, opened 2019-08-23
Import JNIUtils from libjitsi to jitsi-utils.

JonathanLennox:
This will allow JNIUtils to be used without depending on libjitsi.  (In particular I want this for the jitsi-srtp I'm developing.) 

This adds a dependency from jitsi-utils to jna, for its library loading code.

### DIFF
--- a/src/main/java/org/jitsi/utils/JNIUtils.java
+++ b/src/main/java/org/jitsi/utils/JNIUtils.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright @ 2015 Atlassian Pty Ltd
+ *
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/utils/JNIUtils.java
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright @ 2015 Atlassian Pty Ltd
```
bgrozev: Can we update the license header?

JonathanLennox: Do you have a template for what they should now be?

PR #11, opened 2019-08-27
Logger v2

bbaldino:
I think this is looking pretty good now.  I've got some branches in JMT and JVB where I started using it to see how it felt and work out some kinks (I'll push at least the JMT one somewhere so you can see some of the usage).

### DIFF
--- a/src/main/java/org/jitsi/utils/logging2/ContextLogRecord.java
+++ b/src/main/java/org/jitsi/utils/logging2/ContextLogRecord.java
@@ -1,0 +1,25 @@
+package org.jitsi.utils.logging2;
+
+import java.util.logging.*;
+
+/**
+ * {@link ContextLogRecord} extends {@link LogRecord} and adds
+ * a 'context' String.  The reason it's done this way and the context
+ * is not added to the log message itself is so that, in the log formatter,
+ * we can place this context elsewhere (notably before the class and
+ * method names) in the final log message.
+ */
+public class ContextLogRecord extends LogRecord
+{
+    protected final String context;
+    public ContextLogRecord(Level level, String msg, String context)
+    {
+        super(level, msg);
+        this.context = context;
+    }
+
+    public String getContext()
+    {
+        return context;
+    }
+}
--- a/src/main/java/org/jitsi/utils/logging2/ContextLogRecord.java
+++ b/src/main/java/org/jitsi/utils/logging2/ContextLogRecord.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright @ 2018 - present 8x8, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


src/main/java/org/jitsi/utils/logging2/ContextLogRecord.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * {@link ContextLogRecord} extends {@link LogRecord} and adds
+ * a 'context' String.  The reason it's done this way and the context
+ * is not added to the log message itself is so that, in the log formatter,
+ * we can place this context elsewhere (notably before the class and
+ * method names) in the final log message.
+ */
+public class ContextLogRecord extends LogRecord
+{
+    protected final String context;
```
bgrozev: Can this be private?

src/main/java/org/jitsi/utils/logging2/JitsiLogFormatter.java
```diff
@@ -0,0 +1,243 @@
...
+        String cname = JitsiLogFormatter.class.getName();
+        timestampDisabled = Boolean.parseBoolean(
+                manager.getProperty(cname + DISABLE_TIMESTAMP_PROPERTY));
+    }
+
+    /**
+     * Load the programname property to be used in logs to identify Jitsi-based
+     * application which produced the logs
+     */
+    private static void loadProgramNameProperty()
```
bgrozev: Do we still need the program name? I would like to unset it in more cases for the bridge because it is (slightly) easier to read logs without it.

src/main/java/org/jitsi/utils/logging2/Logger.java
```diff
@@ -0,0 +1,182 @@
...
+public interface Logger
+{
+    /**
+     * Create a 'child' logger which derives from this one.  The child logger
+     * will share the same log level setting as this one and its
+     * {@link LogContext} (given here) will inherit all the context
+     * from this logger.
+     * @return
+     */
+    Logger createChildLogger(String name, Map<String, String> context);
```
bgrozev: What does the two-argument method do?

src/main/java/org/jitsi/utils/logging2/Logger.java
```diff
@@ -0,0 +1,182 @@
...
+     * <p>
+     * @param msg   The message to log
+     * @param t  Throwable associated with log message.
+     */
+    void error(Object msg, Throwable t);
+
+    /**
+     * Set logging level for all handlers to ERROR
+     */
+    void setLevelError();
```
bgrozev: Do you think we need all these, or were they just present in the original? I'd remove them and use `setLevel(Level)`

src/main/java/org/jitsi/utils/logging2/Logger.java
bgrozev: I meant it needs a doc because it's not obvious (since this doc says it inherits the context of the parent, but there's also a context passed in)

src/main/java/org/jitsi/utils/logging2/Logger.java
bgrozev: Up to you

src/test/java/org/jitsi/utils/logging2/LoggerImplTest.java
```diff
@@ -0,0 +1,212 @@
+package org.jitsi.utils.logging2;
```
bgrozev: Missing headers

PR #14, opened 2019-09-04
add map creation helpers

bbaldino:


src/main/java/org/jitsi/utils/collections/JMap.java
```diff
@@ -0,0 +1,94 @@
...
+ * Note that both styles--one taking a vararg of Entries and another
+ * taking individual arguments--are in Java 9 and are provided here
+ * for convenience.
+ */
+public class JMap
+{
+    /**
+     * Helper function to create a Map from a list of entries
+     *
+     * NOTE: without @SafeVarargs, we get a complaint about heap
```
JonathanLennox: The Note about @SafeVarargs shouldn't be part of the Javadoc I don't think, unless it's relevant to the user rather than the implementor?

src/main/java/org/jitsi/utils/collections/JMap.java
```diff
@@ -0,0 +1,94 @@
...
+        Map<K, V> map = new HashMap<>();
+        for (Map.Entry<K, V> entry : entries)
+        {
+            map.put(entry.getKey(), entry.getValue());
+        }
+        return map;
+    }
+
+    /**
+     * @param k1
```
JonathanLennox: This Javadoc is missing its description, and the @param values are sort of sparse.

src/main/java/org/jitsi/utils/collections/JMap.java
```diff
@@ -0,0 +1,94 @@
...
+     * @return
+     */
+    public static <K, V> Map<K, V> of(K k1, V v1)
+    {
+        Map<K, V> map = new HashMap<>();
+        map.put(k1, v1);
+        return map;
+    }
+
+    public static <K, V> Map<K, V> of(K k1, V v1, K k2, V v2)
```
JonathanLennox: The rest of these functions need Javadocs entirely.

src/test/java/org/jitsi/utils/collections/JMapTest.java
```diff
@@ -0,0 +1,51 @@
...
+        assertEquals(3, data.size());
+        assertEquals("1", data.get("one"));
+        assertEquals("2", data.get("two"));
+        assertEquals("3", data.get("three"));
+    }
+
+    @Test
+    public void testArgsMapOf()
+    {
+        java.util.Map data = JMap.of("one", "1", "two", "2", "three", "3");
```
JonathanLennox: Should probably have tests of all three variants of JMap.of.

src/test/java/org/jitsi/utils/collections/JMapTest.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.jitsi.utils.collections.JMap.entry;
+import static org.junit.Assert.*;
+
+public class JMapTest
+{
+
+    @Test
+    public void testMapOf()
+    {
+        java.util.Map data = JMap.ofEntries(
```
JonathanLennox: You don't explicitly declare the K and V of the Map here -- do we want to test the case where you have that?

src/main/java/org/jitsi/utils/collections/MapEntry.java
```diff
@@ -0,0 +1,57 @@
...
+    protected final K key;
+    protected V value;
+
+    public MapEntry(K key, V value)
+    {
+        this.key = key;
+        this.value = value;
+    }
+
+    @Override
```
JonathanLennox: Add {@inheritDoc} on all these methods?

src/main/java/org/jitsi/utils/collections/JMap.java
bbaldino: i was worried about whether or not putting a comment below the javadoc comment would 'disassociate' the javadoc comment from the method itself for things trying to parse it (we don't actually do that, so i don't know for sure and couldn't test it).

src/main/java/org/jitsi/utils/collections/JMap.java
bbaldino: fixed

src/main/java/org/jitsi/utils/collections/JMap.java
bbaldino: fixed

src/test/java/org/jitsi/utils/collections/JMapTest.java
bbaldino: fixed

src/test/java/org/jitsi/utils/collections/JMapTest.java
bbaldino: i'm not sure there's a distinction between the cases, actually.  i think the types are derived from the given entries.

bbaldino: at least,  for the purpose of the test.

src/main/java/org/jitsi/utils/collections/MapEntry.java
bbaldino: fixed

PR #15, opened 2019-09-04
copy over xml/os utils from libjitsi

bbaldino:


bbaldino:
decided not to move this stuff over

PR #21, opened 2019-09-19
Support adding log context after LogContext creation

bbaldino:
There's an interaction in jvb with ice4j where we need this, and although it's possible that we may be able to work around not having it some other way, I did find in a couple places when updating the logger that it would be useful, so thought I would give it a try.

The last commit in this PR tweaks the implementation to enforce the use of immutable maps (really annoying Java doesn't expose that type)--but I had to bring in a guava to do so.  I really wanted to enforce this so we don't have to worry about any concurrency issues, so I thought it was worth it (and, to be honest, I've run into many scenarios where guava had a class I could've used); but we can discuss.

### DIFF
--- a/src/main/java/org/jitsi/utils/logging2/LogContext.java
+++ b/src/main/java/org/jitsi/utils/logging2/LogContext.java
@@ -20,6 +20,7 @@ import com.google.common.collect.*;
 import org.jetbrains.annotations.*;
 import org.jitsi.utils.collections.*;
 
+import java.lang.ref.*;
 import java.util.*;
 import java.util.stream.*;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/utils/logging2/LogContext.java
```diff
@@ -16,53 +16,155 @@
...
     {
-        if (context.isEmpty()) {
+        this.context = ImmutableMap.copyOf(context);
+        this.parentContext = parentContext;
+        updateFormattedContext();
+    }
+
+    protected void updateFormattedContext()
+    {
+        this.formattedContext = formatContext(combineMaps(parentContext, context));
```
JonathanLennox: To make sure I understand the Java thread model, this assignment is atomic, so there aren't any multi-threading issues here?  (It's a data race whether other threads get the old or new context, but that's not a problem.)

bbaldino: right

bbaldino:
jenkins, test this please

PR #23, opened 2019-10-08
Config utils

bbaldino:
Helper/base classes for reading values from configuration sources, as well as some tools for validating configuration.  Enables the following use cases/features:

#### A backwards compatible method of transitioning to a new configuration library/file/paradigm

This code abstracts away the source of a configuration property and allows a 'fall through' lookup to check in multiple places, including the property being under a different name in each place and/or being modeled as a different type.  

#### A forwards-compatible way of defining configuration properties

The same things which let us easily interoperate with an old config also make it easy to make changes to config moving forward.  For example, changing the name of a property or the type we use to model a property.  We can easily continue to support the 'old' value for as long as we need while the code can use the new format.

#### Ease the transition for other users/deployers of Jitsi applications

We can detect when old properties are still being used and log warnings with deprecation instructions so that we can phase out old configuration properties, giving users ample warning to make changes.  This way we don't have to support old config properties forever/be bound to an old format because changing things will break backwards compatibility.

#### Detect common configuration issues

In its current state, the config validation code will detect properties which have been defined but are marked as obsolete, as well as property names which are defined but are not read by any class (to detect things like typos)


jitsi-utils/src/main/java/org/jitsi/utils/config/README.md
```diff
@@ -0,0 +1,133 @@
...
+            .readEveryTime()
+            <...snip: not found strategy sections below>
+        );
+    }
+}
+```
+And now, the next time we call `.get()` we'll get a new value if the suppliers return a different result than they did before.  To read a property only once (when the property instance is created), use `.readOnce()` instead.
+
+#### Not found strategy
+Generally, configuration properties must be defined (this wrapper defines no method to supply a default value) so if the code looks for a property and it isn't found, a `ConfigPropertyNotFound` exception is thrown.  Sometimes, though, an optional property is desired.  The 'not found' strategy defines what happens when none of the suppliers find a value for this property.  If we want to require a property, we'd use: `.throwIfNotFound()`:
```
JonathanLennox: It might be worth clarifying that it's up to the configuration supplier to provide the default value.

bbaldino: would this change address what you're saying? want to make sure i'm understanding you right.

> Generally, configuration properties must be defined (this wrapper defines no method to supply a default value), meaning at least one supplier must provide a value for a given property.  If no suppliers produce a value, a ConfigPropertyNotFound exception is thrown. Sometimes, though, an optional property is desired. The 'not found' strategy defines what happens when none of the suppliers find a value for this property. If we want to require a property, we'd use: .throwIfNotFound()

JonathanLennox: Yeah, that seems good.

jitsi-utils/src/main/java/org/jitsi/utils/config/AbstractConfigProperty.java
```diff
@@ -0,0 +1,72 @@
...
+public abstract class AbstractConfigProperty<T> implements ConfigProperty<T>
+{
+    protected final List<Supplier<T>> configValueSuppliers;
+    protected final ReadFrequencyStrategy<T> readFrequencyStrategy;
+    protected final PropNotFoundStrategy<T> propNotFoundStrategy;
+
+    public AbstractConfigProperty(PropertyConfig<T> builder)
+    {
+        builder.validate();
+        this.configValueSuppliers = builder.propValueSuppliers;
```
bgrozev: Nit: I'm a little confused by the prefixes of the names here. Is there a difference between a "config" value supplier and a "prop" value supplier? Assuming there isn't, I think it'd be cleaner to just use `valueSuppliers` in both places.

jitsi-utils/src/main/java/org/jitsi/utils/config/PropertyConfig.java
```diff
@@ -0,0 +1,103 @@
...
+    public PropertyConfig<PropValueType> returnNullIfNotFound()
+    {
+        propNotFoundStrategy = new ReturnNullIfNotFoundStrategy<>();
+
+        return this;
+    }
+
+    public void validate()
+    {
+        if (readFrequencyStrategyCreator == null || propNotFoundStrategy == null)
```
bgrozev: This can be simplified to:
```
if (readFrequencyStrategyCreator == null)
{
    throw new ...
}
if (propNotFoundStrategy == null)
{
    throw new ...
}
```

jitsi-utils/src/main/java/org/jitsi/utils/config/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+    public static void main(String[] args)
+    {
+        new Webserver(webserverPort.get());
+    }
+}
+```
+
+#### Read frequency strategy
+Some configuration values we want to be able to update if we re-load the config in order to see new values.  To get this behavior, we apply a read frequency strategy to our property definition.  So, if we want the value return from `.get()` to update if the config is reloaded and the value has changed, our property definition would become:
```
bgrozev: Nit: s/return/returned/

jitsi-utils/src/main/java/org/jitsi/utils/config/PropertyConfig.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+public class PropertyConfig<PropValueType>
+{
+    /**
+     * The ReadFrequencyStrategy constructor requires a supplier which will give the property's
+     * value, but we don't have that here (it comes from {@link AbstractConfigProperty}.
+     * We'll pass this method to the {@link org.jitsi.utils.config.AbstractConfigProperty}
+     * constructor so that it can pass the provider and instantiate the ReadFrequencyStrategy
+     */
+    protected Function<Supplier<PropValueType>, ReadFrequencyStrategy<PropValueType>> readFrequencyStrategyCreator = null;
```
bgrozev: Nit: Elsewhere in this PR you use `T` for the type, and it would be nice if it's consistent. I'm so used to single letters for abstract types that `PropValueType` reads like a concrete class, so I'd prefer `T`.

jitsi-utils/src/main/java/org/jitsi/utils/config/AbstractConfigProperty.java
bbaldino: thanks, i agree.  fixed.

jitsi-utils/src/main/java/org/jitsi/utils/config/PropertyConfig.java
bbaldino: good point :) fixed

jitsi-utils/src/main/java/org/jitsi/utils/config/README.md
bbaldino: fixed

jitsi-utils/src/main/java/org/jitsi/utils/config/PropertyConfig.java
bbaldino: weird, i grepped for instances of `T` and thought i changed them all to `PropValueType`.  i went back and forth on this, though (between `T` and `PropValueType`)--i'll change back to T.  i do find `T` can be confusing on its own, but it is more common so that's probably better.

PR #24, opened 2019-10-08
Milli formatter

JonathanLennox:
Add function to format (times like) Instant and Duration as full-precision floating-point millis.

### DIFF
--- a/src/main/java/org/jitsi/utils/TimeUtils.java
+++ b/src/main/java/org/jitsi/utils/TimeUtils.java
@@ -208,14 +208,17 @@ public class TimeUtils
         int millis = nanos / 1_000_000;
         int nanosPerMilli = nanos % 1_000_000;
 
-        if (secs != 0) {
+        if (secs != 0)
+        {
             builder.append(secs);
             builder.append(trailingMilliFormat.format(millis));
         }
-        else {
+        else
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/utils/TimeUtils.java
```diff
@@ -175,4 +177,48 @@ public static long getLsw(long ntpTime)
...
+        {
+            secs = -secs - 1;
+            nanos = 1_000_000_000 - nanos;
+            builder.append('-');
+        }
+
+        int millis = nanos / 1_000_000;
+        int nanosPerMilli = nanos % 1_000_000;
+
+        if (secs != 0) {
```
bbaldino: nit: braces in here should be on a newline

src/test/java/org/jitsi/utils/TimeUtilsTest.java
```diff
@@ -0,0 +1,38 @@
+package org.jitsi.utils;
+
+import org.junit.*;
+
+import static org.junit.Assert.*;
+
+public class TimeUtilsTest
+{
+    @Test
+    public void testTimestampFormats() {
```
bbaldino: nit: brace

PR #27, opened 2019-11-08
Configk

bbaldino:
NOTE: this should be considered a draft PR for some more discussion...I accidentally hit submit before changing the type.

This iteration bites the bullet and implements things in kotlin, which I think gives us a lot more options for flexible syntax in how we want to be able to define properties.   As it stands, I'd say this code lays out an un-opinionated foundation; I'd like to lay out some options here and we can discuss the direction and then add code which would make things more opinionated  as  to how properties are defined.

Some things I've played with with a bit of code on top of what's here (some scattered throughout files):

A DSL-style which would allow properties defined like:
```
class LongProperty : ConfigProperty<Long> by property({
    attributes {
        readOnce()
    }
    search(
        "legacyPropName" from legacyConfig() asType(Long::class),
        "newPropName" from newConfig() asType(Duration::class) convertedBy { it.toMillis() }
    )
})
```

A helper parent class for properties which fall under these conditions:
1. Old prop name in old config file
1. New prop name in new config file
1. No type conversions or anything fancy

```
abstract class DefaultConfig<T : Any>(
    clazz: KClass<T>,
    attributes: ConfigPropertyAttributes,
    legacyPropName: String,
    newPropName: String
) : FallbackConfigProperty<T>() {

    private val legacyProp = PropSource(
        clazz,
        attributes = attributes,
        path = legacyPropName,
        configSource = legacyConfig()
    )
    private val newProp = PropSource(
        clazz,
        attributes = attributes,
        path = newPropName,
        configSource = newConfig()
    )

    override val propertyPriority: List<ConfigResult<T>>
        get() = listOf(legacyProp.result, newProp.result)
}
```

Allows us to define a simple property like this:
```
class Simple : DefaultConfig<Int>(Int::class, readOnce(), "legacyName", "newName")
```

I've also played with using property delegates for the properties:
```
class NewHealthTimeout : FallbackConfigProperty<Long>() {
        private val legacyProperty: ConfigResult<Long> by legacyConfigProperty(attributes, legacyPropertyName)
        private val property: ConfigResult<Long> by configProperty<Duration>(attributes, propertyName).convertedBy { it.toMillis() }

        override val propertyPriority: List<ConfigResult<Long>> = listOf(legacyProperty, property)

        companion object {
            private val attributes = readOnce()
            private const val legacyPropertyName = "org.jitsi.videobridge.health.TIMEOUT"
            private const val propertyName = "videobridge.health.timeout"

            val singleton = NewHealthTimeout()
        }
    }
```
The way this is done also makes it easy to directly use the underlying config library's more complex types to do more advanced transformations.  In this example, `ConfigObject` is a `typesafe.config`  type.  We do this by having config libraries dynamically register themselves as providers of 'getters' of certain types, so we don't have to define a single interface with `getInt`, `getBoolean`, etc. (which would also make it impossible to use complex types from specific libraries).  Instead, you request a property value type and, if a provider has registered support for it, you'll get it, otherwise  we'll throw.
```
class ComplexProp :  ConfigProperty<List<Foo>> {
        private val property: ConfigResult<List<Foo>> by configProperty<ConfigObject>(readOnce(), "videobridge.health").convertedBy {
            it.entries.stream()
                    .map { entry -> entry.key }
                    .map(::Foo)
                    .toList()
        }

        override val value: List<Foo>
            get() = property.getOrThrow()

        companion object {
            val singleton = ComplexProp()
        }
    }
```

Note that many of the above classes wouldn't be defined here, they'd be in JVB where it has awareness of old/new config, but using them as examples of how the framework here would be leveraged.


bbaldino:
Made this to hopefully give a better high-level picture
image

The more I'm playing with it, the more I'm thinking we don't really need the property delegates and can just use this style:

```
// A property which has to transform a Duration retrieved from the new 
// config to a Long for backwards compatibility
class TransformingProperty : FallbackConfigProperty<Long>() {
    private val attributes = readOnce()
    val legacyProp = PropValueRetriever.fromConfig(
        Long::class,
        attributes = attributes,
        configKey = "legacyName",
        configSource = legacyConfig()
    )
    val newProp = PropValueRetriever.fromConfig(
        Duration::class,
        attributes = attributes,
        configKey = "newName",
        configSource = newConfig()
    ).convertedBy { it.toMillis() }

    override val propertyPriority: List<ConfigResult<Long>>
        get() = listOf(legacyProp.result, newProp.result)
}
```

The common case of a simple property would still look like:
```
class Simple : DefaultConfig<Int>(Int::class, readOnce(), legacyPropName, newPropName) {
    companion object {
        const val legacyPropName = "legacyName"
        const val newPropName = "newName"
    }
}
```
(The only reason to put the key names as static constants, by the way, is so that we can scrape them for validation.  Maybe can find another way to do that and we could have a one-line definition for simple properties).

There are ways we could improve things here; namely not having to 'repeat' ourselves with the attributes (defining them and then passing them into each property) and being able to just define the properties in their priority order rather than having another method for that.  Going down that path I think leads us to something like the DSL-style:
```
class TransformingProperty : ConfigProperty<Long> by property({
    attributes {
        readOnce()
    }
    search(
        "legacyName" from legacyConfig() asType(Long::class),
        "newName" from newConfig() asType(Duration::class) convertedBy { it.toMillis() }
    )
})
```
which eliminates the drawbacks mentioned above, but is perhaps less declarative.  It'd basically just be a helper to generate the class above, so it wouldn't be a fundamental change.

bbaldino:
Ok, cleaned up the code and a couple other bigger changes (in the implementation):
1. Fixed a bug that was in the old code where we wouldn't properly re-read values for properties configured to be re-read every time
2. Got rid of the SPI thing.  It was an experiment but ended up being a real plain (especially for testing) and I came up with a new way to achieve the same thing that's much simpler.

I included an 'examples' package (which we don't have to merge) which shows a bunch of example property definitions as well as an example `ConfigSource`.  I'll also re-do some of the props in JVB to give more concrete examples there.

PS don't bother going commit-by-commit...it's just a long, twisted forest path of various experimentation.

bbaldino:
Forgot another change: 
3. Instead of searching for all properties via reflection, we can now mark the property itself as deprecated and log whenever it's accessed (or we can change this to just check when initialized).  This also means we don't have to keep the property names in constants somewhere, though it does mean we can no longer validate the scenario where a config property is defined in a config file, but no property class actually reads it.  Previously, we brute-force checked this by iterating over all keys in the config files and then checked all property sub-classes to check if it defined any key names that matched.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/configk/ConfigResult.kt
```diff
@@ -0,0 +1,81 @@
...
+        is ConfigResult.PropertyFound -> this.value
+    }
+}
+
+fun<T : Any> ConfigResult<T>.isFound(): Boolean =
+    this is ConfigResult.PropertyFound
+
+fun<T : Any> ConfigResult<T>.getOrElse(onFailure: (exception: Throwable) -> T): T {
+    return when (this) {
+        is ConfigResult.PropertyNotFound -> throw this.exception
```
bgrozev: Should this call `onFailure` instead of throwing?

bbaldino: yes! nice catch.  i didn't end up using this...so will either fix it or just remove it.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/configk/ConfigResult.kt
```diff
@@ -0,0 +1,81 @@
...
+        is ConfigResult.PropertyFound -> this.value
+    }
+}
+
+fun<T : Any, R> ConfigResult<T>.fold(
```
bbaldino: didn't end up needing this either, so will remove it

jitsi-utils-kotlin/README.md
```diff
@@ -0,0 +1,225 @@
...
+  fromConfig(myConfigSource)
+}
+```
+And we want to start using it as a `Duration` in the code.  We can change the property definition to convert it:
+```
+val timeout = property<Duration> {
+  name("myapp.timeout")
+  readOnce()
+  fromConfig(myConfigSource)
+  retrievedAs<Long>() convertedBy { Duration.ofMillis(it) }
```
bgrozev: This is really cool

PR #29, opened 2019-11-25
Migrate config service interface to jitsi-utils (from libjitsi)

bbaldino:


jitsi-utils/src/main/java/org/jitsi/service/README.md
```diff
@@ -0,0 +1,7 @@
+This package lives in a non-standard location for `jitsi-utils` 
+(`org.jitsi.service` as opposed nested beneath `org.jitsi.utils` like 
+everything else), because the `ConfigurationService` definition was migrated 
+from `libjitsi`, and the `libjitsi` bundle activator relies on implementations 
+having the same package names as their service definitions, but with`service` 
+swapped for `impl`.  In an effort to minimize the change there, the `service` 
+packages lives at the top level in `jitsi-utils`.
```
JonathanLennox: I wonder if this readme should also say something to the effect that this API is intended only for legacy configs, and new code should use the new config APIs?

bbaldino: yeah, i wondered about putting some deprecation notice on the interface itself, as well, but i wasn't sure if we wanted to go that far with it yet.

PR #30, opened 2019-11-26
move passwordutil, debugPrintSystemProperties and masks to jitsi-utils

bbaldino:
Jicoco was accessing `ConfigurationServiceImpl` directly to set the masks so that `debugPrintSystemProperties` masked out the passwords.  Now that the `ConfigurationService` interface is moved to `jitsi-utils`, if we fix this usage then we can break the `jicoco` -> `libjitsi` dependency.

### DIFF
--- a/jitsi-utils/src/main/java/org/jitsi/utils/ConfigUtils.java
+++ b/jitsi-utils/src/main/java/org/jitsi/utils/ConfigUtils.java
@@ -18,6 +18,8 @@ package org.jitsi.utils;
 import org.jitsi.service.configuration.*;
 
 import java.io.*;
+import java.util.*;
+import java.util.regex.*;
 
 /**
  * @author George Politis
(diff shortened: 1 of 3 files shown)


jitsi-utils/src/main/java/org/jitsi/utils/ConfigUtils.java
```diff
@@ -255,4 +257,66 @@ public static String getString(ConfigurationService cfg,
...
+    public static String PASSWORD_SYS_PROPS;
+
+    /**
+     * Goes over all system properties and outputs their names and values for
+     * debug purposes. The method has no effect if the logger is at a log level
+     * other than DEBUG or TRACE (FINE or FINEST).
+     * * Changed that system properties are printed in INFO level and this way
+     *   they are included in the beginning of every users log file.
+     */
+    public static String debugPrintSystemProperties()
```
bgrozev: Can we clean this up a little? The name and description are confusing.

PR #31, opened 2019-12-04
Fix deprecation logs

bbaldino:


bbaldino:
If you've got a bette idea on how to do the "+ 1" logic in the tests, I'd be  interested in hearing it...this way doesn't feel great.

bbaldino:
Maybe:
```
   //NOTE: The "+ 1" in read-every-time properties when checking invocation
    // counts it to take into account the fact that we retrieve the value
    // at creation time for all props in order to print a deprecation message
    // (since normally those props won't be accessed)
    private fun expectedAccessCount(explicitAccessCount: Int): Int =
        explicitAccessCount + 1
```

and then the  test-site would be:
```
should("query the value every time") {
    repeat (5) { property.value }
    legacyConfig.numGetsCalled shouldBe expectedAccessCount(5)
    newConfig.numGetsCalled shouldBe expectedAccessCount(0)
}
```

JonathanLennox:
Is there any way that the deprecation check could bypass whatever logic increments `numGetsCalled`?

bbaldino:
> Is there any way that the deprecation check could bypass whatever logic increments `numGetsCalled`?

No, unfortunately not because it needs to check whether or not the property was actually found, so it calls the same logic any normal access does.

bbaldino:
I tweaked the test to use the helper, I think with the comment and logic in a central place it's better. (But still open to other ideas)

PR #39, opened 2020-01-23
Add joinToRangedString function.

JonathanLennox:
Print partially-consecutive sets of integers as (e.g.) "1-5, 7, 9-11".

### DIFF
--- a/jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/RangedString.kt
+++ b/jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/RangedString.kt
@@ -1,0 +1,101 @@
+/*
+ * Copyright @ 2018 - Present, 8x8 Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jitsi.utils
+
+/** Format a sequence of ints as a ranged string, e.g. "1, 3-8, 9-10" */
+/* TODO: it'd be nice to support this for any integer type but I don't know
+    if there's a good way to represent this.  Unfortunately interface [Number]
+    doesn't have [operator fun plus], so we can't use it to test
+    element == previous + 1.
+ */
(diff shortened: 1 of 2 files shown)


jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/RangedString.kt
```diff
@@ -0,0 +1,101 @@
...
+}
+
+fun Iterable<Int>.joinToRangedString(
+    separator: CharSequence = ", ",
+    rangeSeparator: CharSequence = "-",
+    prefix: CharSequence = "",
+    postfix: CharSequence = "",
+    rangeLimit: Int = -1,
+    truncated: CharSequence = "..."
+): String = this.iterator().joinToRangedString(separator = separator, rangeSeparator = rangeSeparator, prefix = prefix, postfix = postfix, rangeLimit = rangeLimit, truncated = truncated)
```
bbaldino: Why the named args for these?  It doesn't seem like they are necessary.

bbaldino:
So, I played with this a little bit (mostly for fun) and simplified the main loop logic a bit by calling out a couple corner cases at the top.  Whether or not this is better may be subjective, but I'll put it here in case there's anything you find appealing:
```
fun Iterator<Int>.joinToRangedString(
    separator: CharSequence = ", ",
    rangeSeparator: CharSequence = "-",
    prefix: CharSequence = "",
    postfix: CharSequence = "",
    rangeLimit: Int = -1,
    truncated: CharSequence = "..."
): String = with(StringBuilder()) {
    append(prefix)
    if (!hasNext()) {
        return append(postfix).toString()
    }
    if (rangeLimit == 0) {
        return append(truncated).toString()
    }

    var rangeCount = 0
    fun canAddElement(): Boolean {
        return when {
            rangeLimit < 0 -> true
            rangeCount < rangeLimit -> true
            else -> false
        }
    }

    var inRange = false
    var previous = next()
    append(previous)

    loop@ while (hasNext()) {
        val element = next()
        when (element) {
            previous + 1 -> inRange = true
            else -> {
                if (inRange) {
                    append(rangeSeparator).append(previous)
                    inRange = false
                }
                append(separator)
                rangeCount++
                if (canAddElement()) {
                   append(element)
                } else {
                    append(truncated)
                    break@loop
                }
            }
        }
        previous = element
    }

    if (inRange) {
        append(rangeSeparator).append(previous)
    }
    append(postfix)

    toString()
}
```

bbaldino:
Also, based on our discussion, I think some docs for the semantics of `rangeLimit` would be useful.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/RangedString.kt
JonathanLennox: Because they're all `CharSequence`s, so getting them out of order as non-named args would be really easy?

PR #42, opened 2020-01-31
WIP: collect packet queue stats on queue linger times

JonathanLennox:
~~Unfortunately this implementation has a 5-10% performance overhead (measured with `PacketQueueBenchmarkTests`) even when statistics are disabled, presumably due to the allocations of the `QueueElement` objects, so any suggestions as to how to avoid this would be welcome.~~

This has now been changed to put statistics gathering into a subclass - this saves the overhead, at the expense of requiring a factory to create `PacketQueue`.  Comments on this welcome too.

(Note that the actual calculations of the queue linger time is not yet implemented, just the data gathering.)


JonathanLennox:
Closed in favor of https://github.com/jitsi/jitsi-utils/pull/47

PR #43, opened 2020-01-31
Add Outcome helper

bbaldino:
From the docs: 
```
[Outcome] models the outcome of some process which is not yet known.  Once 
known, it will not change.  It allows subscribing to both the success and
failure of the outcome.  Subscribers are notified once the outcome has
been reached, or instantly if the outcome has already been reached.

Until an outcome is reached, the state is 'unknown' (it has neither succeeded
nor failed).  This can be tested explicitly with [isKnown].
```
I'm playing (yak-shaving) with some stuff around transport in JVB, and found this mechanism really useful for tracking things like ICE state/DTLS handshake state.


bbaldino:
I went about the transport refactoring another way (https://github.com/jitsi/jitsi-videobridge/pull/1104), and I'm not sure this helper made much sense (especially since we'll want to support ICE restarts).

PR #44, opened 2020-02-05
Spotbugs

bgrozev:


jitsi-utils/src/main/java/org/jitsi/utils/stats/RateStatistics.java
```diff
@@ -102,45 +104,42 @@ private synchronized void eraseOld(long nowMs)
         oldestTime = newOldestTime;
     }
 
-    public long getRate()
+    public synchronized long getRate()
```
bbaldino: We use this class in some critical-path areas, right?  We may want to do some testing before adding a bunch of `synchronized`.

jitsi-utils/src/test/java/org/jitsi/utils/BasicByteArrayBuffer.java
```diff
@@ -1,5 +1,10 @@
 package org.jitsi.utils;
 
+import edu.umd.cs.findbugs.annotations.*;
+
+@SuppressFBWarnings(
+    value = "EI_EXPOSE_REP",
+    justification = "It is the nature of this class to expose 'buffer.'")
```
bbaldino: Is this firing because we have a private member that we return via the getter?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/config/FallbackProperty.kt
```diff
@@ -24,7 +24,7 @@ import org.jitsi.utils.config.strategy.getReadStrategy
  * to find the value
  */
 abstract class FallbackProperty <T : Any>(
-    vararg val attrs: ConfigPropertyAttributes<T>
+    vararg attrs: ConfigPropertyAttributes<T>
```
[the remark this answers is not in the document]
bgrozev: It's not used right now, but if it is indented to be exposed I can revert the change. Do you know what the best way to mark it as intentionally exposed? Supress the warning or something else?

jitsi-utils/src/main/java/org/jitsi/utils/concurrent/ExecutorUtils.java
```diff
@@ -71,11 +73,7 @@ public Thread newThread(Runnable r)
                                 if ((baseName != null)
                                         && (baseName.length() != 0))
                                 {
-                                    String name = t.getName();
-
-                                    if (name == null)
-                                        name = "";
-                                    t.setName(baseName + "-" + name);
+                                    t.setName(baseName + "-" + t.getName());
```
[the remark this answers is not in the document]
bgrozev: Yes, but it was detected as a superfluous check, so I guess we can safely assume `Thread.getName()` never returns null.

jitsi-utils/src/main/java/org/jitsi/utils/stats/RateStatistics.java
bgrozev: All of these methods call `eraseOld`, which is already synchronized, so there are no additional lock aquisitions. If we wanted to optimize to avoid aqcuiring a lock we know we already hold (the impact of which is minimal AFAIK) we could remove some of the `synchronized`, but:
1. We'd have a warning to supress and it makes the code easier to break
2. These are pretty simple cases that the compiler could catch and optimize automatically, so I suspect it does so (I haven't checked)

jitsi-utils/src/test/java/org/jitsi/utils/BasicByteArrayBuffer.java
bgrozev: No, it's the same one we discussed in jitsi-rtp: the class has a `private` field of an array type, which is also exposed publically. Here it it exposed via `getBuffer`, in rtp it was indirectly exposed because it was set externally.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/config/FallbackProperty.kt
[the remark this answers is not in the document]
bbaldino: It may have be intended for future use, but we can remove it as a member for now if it's not used anywhere.

PR #45, opened 2020-02-11
Xmpp delay stats

bgrozev:


jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/AtomicExtensions.kt
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.utils
+
+import java.lang.Long.max
+import java.lang.Long.min
+import java.util.concurrent.atomic.AtomicLong
+
+fun AtomicLong.increaseAndGet(newValue: Long): Long
```
bbaldino: I think these names are confusing.  I'd rather do something clear (albeit perhaps long-winded) like `AtomicLong.setToMaxOfCurrentOr(val: Long)` or something like that?

JonathanLennox:
You seem to have created this PR branch off your spotbugs branch, which makes reviewing it more complicated than necessary.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/AtomicExtensions.kt
bgrozev: I named them similar to `incrementAndGet`, but I can see how that can be confusing since the semantics are slightly different. How about `bumpTo` and `dropTo`?

JonathanLennox: Maybe `minAssign` and `maxAssign` by analogy with `plusAssign` etc?

JonathanLennox:
Looks like this has `ktlint` failures?

PR #47, opened 2020-03-06
Queue stats enhancements

JonathanLennox:
Add stats to PacketQueue, and restructure how its stat listener works.


jitsi-utils/src/main/java/org/jitsi/utils/queue/QueueStatistics.java
```diff
@@ -88,54 +108,65 @@ public JSONObject getStats()
...
     /**
      * Registers the addition of a packet.
-     * @param now the time (in milliseconds since the epoch) at which the
-     * packet was added.
      */
-    public void add(long now)
+    @Override
+    public void added(Object o)
     {
+        long now = System.currentTimeMillis();
```
bbaldino: Can we use a `Clock` instead?  I think at some point we'll want to replace all uses of `System.currentTimeMillis()` as it doesn't work well with tests.

jitsi-utils/src/main/java/org/jitsi/utils/queue/QueueStatistics.java
```diff
@@ -88,54 +108,65 @@ public JSONObject getStats()
...
+        stats.put("added", totalPacketsAdded.sum());
+        stats.put("removed", totalPacketsRemoved.sum());
+        stats.put("dropped", totalPacketsDropped.sum());
         stats.put("add_rate", addRate.getRate(now));
         stats.put("remove_rate", removeRate.getRate(now));
         stats.put("drop_rate", dropRate.getRate(now));
         double duration = (now - firstPacketAddedMs) / 1000d;
         stats.put("duration_s", duration);
-        stats.put("average_remove_rate_pps", totalPacketsRemoved.get() / duration);
+        stats.put("average_remove_rate_pps", totalPacketsRemoved.sum() / duration);
```
bbaldino: Add rate also seems like it'd be important here (so we can compare add rate and remove rate)

bbaldino: Oops, I see that was already there.  What's the difference in units?  Maybe we should make those other stat names more obvious.  If they're not "items per second", what are they? Bps?

jitsi-utils/src/main/java/org/jitsi/utils/queue/PacketQueue.java
```diff
@@ -43,19 +43,10 @@
...
-     * The default value for the {@code enableStatistics} constructor argument.
-     */
-    private static boolean enableStatisticsDefault = false;
-
-    /**
-     * Sets the default value for the {@code enableStatistics} constructor
-     * parameter.
-     *
-     * @param enable the value to set.
+     * Deprecated. Does nothing.
```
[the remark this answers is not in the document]
JonathanLennox: My goal before this is done is to remove it; I just didn't feel like editing the videobridge project yet.

jitsi-utils/src/main/java/org/jitsi/utils/queue/QueueStatistics.java
JonathanLennox: The thing I'm uncertain about with using `Clock` is that at some point I'd like to switch to using `System.nanoTime()` across our system for things that don't actually need wall-clock times, and as far as I know there's no `Clock` defined for that. (We could define one, of course, but it's a little tricky because some aspects of `Clock` assume that it has a 1970 epoch.)

jitsi-utils/src/main/java/org/jitsi/utils/queue/QueueStatistics.java
JonathanLennox: I'm not actually 100% sure - these were all stats that were there previously.

That said, I believe the `getRate()` rates are calculated with `RateStatistics`, so over a 5 second window, whereas `average_remove_rate_pps` is calculated since the beginning of time.  The former is probably more useful...

JonathanLennox:
> I think there is a perf harness for the queue, right? Did you compare before/after these changes?

I added statistics to the PacketQueueBenchmarkTests, and this is what I see on my Mac:

```
CachedThreadPerQueuePool : avg = 475 ms, std_dev = 1 ms
CachedThreadPerQueuePool (With Statistics) : avg = 441 ms, std_dev = 0 ms
FixedSizeCPUBoundPool : avg = 504 ms, std_dev = 6 ms
FixedSizeCPUBoundPool (With Statistics) : avg = 536 ms, std_dev = 7 ms
ForkJoinCPUBoundPool : avg = 495 ms, std_dev = 2 ms
ForkJoinCPUBoundPool (With Statistics) : avg = 531 ms, std_dev = 0 ms
ThreadPerQueuePool : avg = 509 ms, std_dev = 3 ms
ThreadPerQueuePool (With Statistics) : avg = 545 ms, std_dev = 5 ms
```

I don't know why `CachedThreadPerQueuePool` was slower without statistics, which leads me to be somewhat dubious of the test, but otherwise it looks like there's some overhead but not a huge amount.

I'll do a YourKit profile of the JVB as well.

JonathanLennox:
Here's the results of a YourKit profile on the JVB (25 participants, all senders):
Screen Shot 2020-03-09 at 11 50 19 AM

bbaldino:
> Here's the results of a YourKit profile on the JVB (25 participants, all senders):

This is the before or the after?

JonathanLennox:
> This is the before or the after?

This is the after.  `QueueStatistics` wouldn't be enabled (without REST API calls) before.

jitsi-utils/src/main/java/org/jitsi/utils/queue/QueueStatistics.java
```diff
@@ -88,54 +108,65 @@ public JSONObject getStats()
...
+        stats.put("average_queue_size_at_remove", totalSize.sum() / (double)totalPacketsRemoved.sum());
+        stats.put("average_queue_wait_time_ms", totalWait.sum() / (double)totalPacketsRemoved.sum());
```
bgrozev: Cache the result of `totalPacketsRemoved.sum()` in a local variable?

JonathanLennox:
So unfortunately this is pretty computationally expensive:
Screen Shot 2020-03-17 at 3 05 05 PM

Any suggestions on making it cheaper, or which things we can still usefully leave out, are welcome.

JonathanLennox:
With the latest series of patches, the CPU load is looking much more reasonable (same test as before):
Screen Shot 2020-03-20 at 2 57 50 PM

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
```diff
@@ -0,0 +1,227 @@
...
+     * Gets a snapshot of the stats in JSON format.
+     */
+    val stats: OrderedJsonObject
+        get() {
+            val stats = OrderedJsonObject()
+            val now = System.currentTimeMillis()
+            stats["added"] = totalPacketsAdded.sum()
+            stats["removed"] = totalPacketsRemoved.sum()
+            stats["dropped"] = totalPacketsDropped.sum()
+            val duration = (now - firstPacketAddedMs) / 1000.0
```
bbaldino: nit: name `durationSecs` or, even better, use `Instant`s and a real `Duration`! :)

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
```diff
@@ -0,0 +1,227 @@
...
+        if (firstPacketAddedMs < 0) {
+            firstPacketAddedMs = now
+        }
+        totalPacketsAdded.increment()
+    }
+
+    /**
+     * Registers the removal of a packet.
+     */
+    fun removed(now: Long, queueSize: Int, waitTime: Long?) {
```
bbaldino: Looks like we don't use `now` here?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
```diff
@@ -0,0 +1,227 @@
...
+        queueLengthStats.addValue(queueSize.toLong())
+        if (waitTime != null) {
+            queueWaitStats?.addValue(waitTime)
+        }
+    }
+
+    /**
+     * Registers that a packet was dropped.
+     */
+    fun dropped(now: Long) {
```
bbaldino: And here?  Were these left over from the more-complete stats?

bbaldino: Also I think it'd be nice to use `Instant` and `Duration` everywhere in here.  If not, let's at least name the variables with units.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
```diff
@@ -0,0 +1,227 @@
...
+
+        /**
+         * The interval for which to calculate rates in milliseconds.
+         */
+        private const val INTERVAL_MS = 5000
+
+        /**
+         * Calculate the statistics buckets for a given queue length
+         */
+        private fun lengthBucketSizes(size: Int): LongArray {
```
bbaldino: nit: `size` -> `queueLength` (or `queueSize`), since that's what it seems to represent?

bbaldino: nit: `lengthBucketSizes` -> `getLengthBucketSizes` ?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
```diff
@@ -0,0 +1,227 @@
...
+    val queueDebugState: JSONObject
+        get() {
+            val d = queue.debugState
+            stats?.let { d["statistics"] = it }
+            return d
+        }
+
+    companion object {
+        var debug = false
+        var trackTimes = false
```
bbaldino: I know these aren't constants, but when reading the code it still kinda feels like maybe they should be in all caps.  I doubt we're consistent on that though (I'm probably not even consistent myself)

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
```diff
@@ -0,0 +1,227 @@
...
+     * The queue being monitored
+     */
+    private val queue: PacketQueue<*>
+) : PacketQueue.Observer {
+    /**
+     * A map of the time when objects were put in the queue
+     */
+    private val insertionTime = if (trackTimes) Collections.synchronizedMap(IdentityHashMap<Any, Long>()) else null
+
+    private val specific = if (debug) QueueStatisticsInternal(queue.capacity()) else null
```
bbaldino: `specific` means "stats for this specific queue"?  How about `localStats` and `globalStats` for these vars?

jitsi-utils/src/test/java/org/jitsi/utils/queue/DummyQueue.java
```diff
@@ -31,13 +31,13 @@
         PacketHandler<Dummy> packetHandler,
```
bbaldino: We have 2 `DummyQueue`s now?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
```diff
@@ -0,0 +1,227 @@
...
+import org.jitsi.utils.OrderedJsonObject
+import org.jitsi.utils.stats.BucketStats
+import org.jitsi.utils.stats.RateStatistics
+import org.json.simple.JSONObject
+import java.util.Collections
+import java.util.concurrent.atomic.LongAdder
+import java.util.IdentityHashMap
+import java.util.concurrent.ConcurrentHashMap
+
+class QueueStatisticsInternal(queueSize: Int) {
```
bbaldino: Probably just confused by the diff, but I'm a little confused by `QueueStatisticsInternal` vs `QueueStatistics`

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
JonathanLennox: Yeah, it was for the old Rate calculation.  I don't need it in the base.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
JonathanLennox: Works for me.

jitsi-utils/src/test/java/org/jitsi/utils/queue/DummyQueue.java
JonathanLennox: One in jitsi-utils and one in jitsi-utils-kotlin, since maven package dependencies don't import their dependencies' test directories.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
JonathanLennox: `QueueStatisticsInternal` is the object that's instantiated as `localStats` and `globalStats`, whereas `QueueStatistics` is the object that actually observes the queue, calculates the current size and (possibly) linger time, and then updates both statistics objects.  Better names welcome....

bbaldino: `QueueStatisticsInternal` -> `QueueStatistics`
`QueueStatistics` -> `QueueObserver` ?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
JonathanLennox: I did this, but it makes setters in Java look like `QueueObserver.Companion.setDEBUG(enabled);` which is pretty ugly.

bbaldino: hm, can you use `@JvmField` here so it's just `QueueObserver.Companion.DEBUG = enabled`?

JonathanLennox: I can, but it triggers Spotbugs complaining 
```
[ERROR] org.jitsi.utils.queue.QueueObserver.DEBUG isn't final but should be [org.jitsi.utils.queue.QueueObserver] In QueueObserver.kt MS_SHOULD_BE_FINAL
[ERROR] org.jitsi.utils.queue.QueueObserver.TRACK_TIMES isn't final but should be [org.jitsi.utils.queue.QueueObserver] In QueueObserver.kt MS_SHOULD_BE_FINAL
```

I'm inclined to just change them back to lower case.

bbaldino: Ugh.  I don't feel like we should have spotbugs doing 'linting' for naming conventions like this...we've already got ktlint for that (which presumably doesn't complain?)

JonathanLennox: It's not enforcing naming conventions -- it's enforcing "don't have static public non-final fields."

(It's safer for primitive types than it is for objects, and boolean is basically completely safe if you don't need synchronization; but it looks like their rule doesn't have an exception for it.)

bbaldino: If so, then it's not complaining about the case, it's complaining about `@JvmField`, no?  Either way, this seems like a reasonable exception.  We could disable the rule for this line/scope.  Part of the problem with SpotBugs is that it's designed for Java, not Kotlin.

jitsi-utils-kotlin/src/test/kotlin/org/jitsi/utils/queue/QueueObserverBenchmarkTest.kt
```diff
@@ -199,7 +199,7 @@ class QueueObserverBenchmarkTest {
         var sumSquares: Long = 0
         for (duration in experimentDuration) {
             val diff = Math.abs(duration.toNanos() - averageNanos)
-            sumSquares = diff * diff
+            sumSquares += diff * diff
```
gpolitis, line 202: stddev and avg calculation seems like a good candidate for a MathUtils class? (or if we don't mind being a little bit more invasive bring Apache Commons Math to the project)?

JonathanLennox:
Closed in favor of https://github.com/jitsi/jitsi-utils/pull/97.

PR #50, opened 2020-03-26
Prefere native library from java.library.path.

mstyura:
This PR change the preference of native libraries lookup.

The PR proposes to always try loading native library from `java.library.path`.
This allow end-user to manually extract library from jar without removing it.

Currently when library extracted into `/tmp/jna/...` it removed immediately after being loaded into JVB.
This is very unsuitable for native debugging. 
For example, `gdb` has no idea about deleted file and can not provide relevant debugging information for loaded library.
```
(gdb) info sharedlibrary
From                To                  Syms Read   Shared Object Library
0x00007ffedf8e83f0  0x00007ffedf8f6a38  Yes (*)     /tmp/jna-1356855001/jna3054194125002259017.tmp
```

With this change user can manually unpack library from jar and it will be picked up by `JNIUtils.loadLibrary`. User can preserve library on disk to facilitate debugging.

### DIFF
--- a/jitsi-utils/src/main/java/org/jitsi/utils/JNIUtils.java
+++ b/jitsi-utils/src/main/java/org/jitsi/utils/JNIUtils.java
@@ -50,10 +50,10 @@ public final class JNIUtils
     {
         try
         {
-            // Always prefer library from java.library.path over unpacked from jar.
-            // It allows end user to manually unpack native libraries and store them
+            // Always prefer libraries from java.library.path over those unpacked from the jar.
+            // This allows end user to manually unpack native libraries and store them
             // in java.library.path to later load via System.loadLibrary.
-            // This allow end-user to preserve native library on disk,
+            // This allows end-users to preserve native libraries on disk,
             // which is necessary for debuggers like gdb to load symbols.
             System.loadLibrary(libname);
             return;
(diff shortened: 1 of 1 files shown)


jitsi-utils/src/main/java/org/jitsi/utils/JNIUtils.java
```diff
@@ -49,13 +49,25 @@ private static void loadLibrary(String libname, Class clazz,
         ClassLoader classLoader)
     {
         try
+        {
+            // Always prefer library from java.library.path over unpacked from jar.
+            // It allows end user to manually unpack native libraries and store them
+            // in java.library.path to later load via System.loadLibrary.
+            // This allow end-user to preserve native library on disk,
+            // which is necessary for debuggers like gdb to load symbols.
+            System.loadLibrary(libname);
```
JonathanLennox: Re-wording the grammar:

```
            // Always prefer libraries from java.library.path over those unpacked from the jar.
            // This allows end user to manually unpack native libraries and store them
            // in java.library.path to later load via System.loadLibrary.
            // This allows end-users to preserve native libraries on disk,
            // which is necessary for debuggers like gdb to load symbols.
```

JonathanLennox:
Jenkins, test this please.

JonathanLennox:
Jenkins, add to whitelist.

PR #53, opened 2020-04-08
call releasePacket for all items in the queue when closing

bbaldino:
@mstyura what do you think of this?

### DIFF
--- a/jitsi-utils/src/main/java/org/jitsi/utils/queue/PacketQueue.java
+++ b/jitsi-utils/src/main/java/org/jitsi/utils/queue/PacketQueue.java
@@ -578,10 +578,6 @@ public abstract class PacketQueue<T>
             {
                 errorHandler.packetHandlingFailed(t);
             }
-            finally
-            {
-                releasePacket(item);
-            }
         }
     }
 }
--- a/jitsi-utils/src/main/java/org/jitsi/utils/queue/PacketQueue.java
+++ b/jitsi-utils/src/main/java/org/jitsi/utils/queue/PacketQueue.java
@@ -421,6 +421,10 @@ public abstract class PacketQueue<T>
                 // threads waiting on queue must stop reading it.
                 queue.notifyAll();
             }
+            T item;
+            while ((item = queue.poll()) != null) {
+                releasePacket(item);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


jitsi-utils/src/main/java/org/jitsi/utils/queue/PacketQueue.java
```diff
@@ -421,6 +421,10 @@ public void close()
                 // threads waiting on queue must stop reading it.
                 queue.notifyAll();
             }
+            T item;
+            while ((item = queue.poll()) != null) {
+                releasePacket(item);
+            }
```
mstyura, line 427: @bbaldino cool! That's probably fixes leaking when queue is used with pooled buffers. I see similar code exists in `doAdd` when packet is dropped when queue is overflowed: 
https://github.com/jitsi/jitsi-utils/blob/75d0df84b6ab1af48a336799a10e6fe0aaf58f62/jitsi-utils/src/main/java/org/jitsi/utils/queue/PacketQueue.java#L292-L305
So taking into account the code quoted above maybe it makes sense also report error handler and statistics that packet is dropped? And maybe it make sense to call `notifyAll` after queue is cleared? What do you think?

bbaldino: > That's probably fixes leaking when queue is used with pooled buffers

Yeah, that's exactly how I got onto looking at this.

> So taking into account the code quoted above maybe it makes sense also report error handler and statistics that packet is dropped?

Actually I intentionally left out a call to `packetDropped` here because I was considering this a non-error scenario--though I guess it could be argued that some use cases would consider this an error case (we wouldn't, though, which is why I omitted it). 

> And maybe it make sense to call notifyAll after queue is cleared?

Do you think this is needed since there's a call to that above?

mstyura: I mean move notify from "before" to "after" clearing the queue. But that's probably a matter of taste. So as long as you've confirmed that calling `drop` and `packetDropped` is intentionally omitted the changes look good to me.

bbaldino: > I mean move notify from "before" to "after" clearing the queue. But that's probably a matter of taste.

Ah ok...yeah that's actually one thing I wanted your input on.  If you think it's fine either way then yeah I figured I'd get anyone waiting on the queue out of the way and then flush the queue.

mstyura: For this particular case I believe there would be no difference between notifying before and after due to `closed` check, but in general I think more correct would be changing state of `queue` (draining all items) and only then notify all possible waiter that state is changed.

bbaldino: But the only case for someone waiting on the queue still would be if the queue is empty anyways, isn't it?

mstyura: Yes, make sense :)

PR #56, opened 2020-04-20
Added ObservableProperty specialization for Int and Long primitives to avoid excessive heap allocations.

mstyura:
During performance testing of `JVB 2` I've noticed that one of the fastest allocating objects are `java.lang.Integer` and `java.lang.Long`.

Big portion of these allocations came from [`RtpPacket`, which used generic `kotlin.properties.ObservableProperty<T, U>`](
https://github.com/jitsi/jitsi-rtp/blob/fcf2de0ab4bb8a13950b07f56a0e22665989596b/src/main/kotlin/org/jitsi/rtp/rtp/RtpPacket.kt#L85-L100)

These excessive allocations could be easily avoided when primitive-specialized types are used. Unfortunately there is no built in primitive types friendly classes in `kotlin.properties`, so I've added one to `jitsi.utils` (by copying and modifying relevant source from Kotlin's stdlib).

An alternative would be proposing primitive types specialization [to Kotlin itself](https://github.com/JetBrains/kotlin/tree/master/libraries/stdlib/src/kotlin/properties), but I will definetely take much longer time, before can actually be used in `RtpPacket`.


bbaldino:
Is this because the generic arg forces a boxing of the type which would otherwise just be passed-through using the native version?

JonathanLennox:
Do you have any results for how much this helps performance?

mstyura:
@bbaldino yes, generics in `JVM` does not work with primitive types, so any `long(java)/Long(kotlin)`, `int(java)/Int(kotlin)` will be boxed - wrapped into `java.lang.Long`, `java.lang.Integer` and put on the heap. 
That's because `JVM` does not know about generic types in runtime, so compiler has to "erase" all generic type information and in runtime every generic type argument is look like "an object reference".
The goal to avoid boxing is why `Java`'s standard library has such strange things like [`IntFunction`](https://docs.oracle.com/javase/8/docs/api/java/util/function/IntFunction.html), [`LongFunction`](https://docs.oracle.com/javase/8/docs/api/java/util/function/LongFunction.html) and why Kotlin has such a "hardcoded" overloads like [`toIntArray`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-int-array.html) and [`toByteArray`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/to-byte-array.html).

mstyura:
@JonathanLennox not jet and actually was not planned to measure how this particular change enhance performance. But I've noticed that for a conference with 6 video peers with LastN = 1 allocation of   `java.lang.Long` was about `~2000` objects per second and `~2300` `java.lang.Integer` objects per second. I consider this a significant allocation rate. Ideally no primitives should go to heap at all :)

mstyura:
Ah, sorry for misleading comment, an actual allocation rate is lower - about ~250 objects per seconds for long and ~100 objects per second for int. I've confused Visual VM "Live Objects" meaning in "delta" mode.

JonathanLennox:
The other piece which is going to have a substantial allocation of primitive types is the various sets and maps indexed by Int and Long - I believe every lookup in one of those classes requires object boxing.

mstyura:
@JonathanLennox yes and this also has to be fixed, I've already noticed there is some `LruCache` allocating lots of `Long`s. This fix I hope is just a starting point on my work on reducing memory allocations.

JonathanLennox:
What I'm wondering is just how costly these boxed allocations are - I've heard that the JVM is "good" at this, but I don't have further information than that.

I'm reluctant to introduce too much new complexity if there isn't a substantial win (for instance, cloning the Kotlin stdlib means tracking any changes/bugfixes there) so it'd be good to see a practical advantage (e.g. fewer GC pauses).

mstyura:
@JonathanLennox yes, allocations are sometimes considered cheap and it mean cheap to allocate - just bump the pointer in a heap, but it's only half of truth, later all excessive allocations are paid during collection. And they are paid during runtime, due to extra level of indirection to follow the poiner, fewer useful data in cache. All of this hard to measure in isolate. I just follow a practical advice that such a heavy-loaded network-intensive application like JVB should do it's best of not allocating when it's possible, especially on hot-paths.
The presence of specialized types for primitives in {{JDK}} is a clear indication that allocations are not cheap for applications which cares about high performance should preferably use specialized types.
An extra evidence that allocations are not cheap is the presence of escape analysis in JVM's JIT, which tries to analyze if it possible to allocate object instance on a stack, rather that on a heap, so the cost of collection is not paid later.

> cloning the Kotlin stdlib means tracking any changes/bugfixes there

In this particular case it's unlikely due to simplicity of the class it will ever changed. But I'll create PR which propose adding primitive types specialization to Kotlin stdlib as well. But as it a long path to make thing fixed I prefer now having "fixed" copy in jitsi-utils.

JonathanLennox:
> so it'd be good to see a practical advantage (e.g. fewer GC pauses).

Or at least how much the allocation rate goes down.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/properties/Interfaces.kt
```diff
@@ -18,20 +18,94 @@ package org.jitsi.utils.properties
...
+/**
+ * Base interface that can be used for implementing property delegates of read-only properties
+ * of int primitive type.
+ *
+ * This is provided only for convenience; you don't have to extend this interface
+ * as long as your property delegate has methods with the same signatures.
+ *
+ * @param T the type of object which owns the delegated property.
+ */
 public interface ReadOnlyIntProperty<in T> {
```
bbaldino, line 30: Shouldn't these be non-generic?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/properties/Delegates.kt
```diff
@@ -18,13 +18,30 @@ package org.jitsi.utils.properties
 
 import kotlin.reflect.KProperty
 
+/**
+ * Property delegates specialized for primitive-types properties.
+ */
 public object Delegates {
+    /**
+     * Returns a property delegate for a read/write property that calls a specified callback function when changed.
```
JonathanLennox, line 26: This should explicitly say it's for ints.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/properties/Interfaces.kt
mstyura: @bbaldino not sure, original generic type has 2 generic parameters, 1 to describe owner type and 1 to describe type of value stored in property. So I've basically [copy-pasted original code from `kotlin.properties`](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/properties/Interfaces.kt) and replaced type argument of property type with change in name of the type.

bbaldino: Ok, I guess we could prove this out by comparing the allocations rates before/after (and also seeing if this spot is causing an allocation).  Did you get a chance to try that?

mstyura: @bbaldino yes, sure. I'll attach screen recording of Visual VM with observation I had before and after changes for a video conference of 10 peers with 640x480 video, simulcast enabled and LastN=1. Does it sound good? I'm was almost done with videos, but was interrupted to [proposing changes to Kotlin library](https://github.com/JetBrains/kotlin/pull/3320).

mstyura: @bbaldino I'm really sorry, I won't be able to provide measurements today. I'm just run out of time today :( and also probably thought of an alternative and probably better idea how to fix an issue.

bbaldino: No problem

mstyura:
Obsoleted in favor of https://github.com/jitsi/jitsi-rtp/pull/79 and https://github.com/jitsi/jitsi-media-transform/pull/259

PR #59, opened 2020-05-21
OSGi manifest, cleanup poms, update dependencies, GH actions build

ibauersachs:
Closes #48

### DIFF
--- a/jitsi-utils-kotlin/pom.xml
+++ b/jitsi-utils-kotlin/pom.xml
@@ -1,148 +1,100 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <modelVersion>4.0.0</modelVersion>
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>org.jitsi</groupId>
-        <artifactId>jitsi-utils-parent</artifactId>
-        <version>1.0-SNAPSHOT</version>
-    </parent>
+  <parent>
+    <groupId>org.jitsi</groupId>
+    <artifactId>jitsi-utils-parent</artifactId>
+    <version>1.0-SNAPSHOT</version>
+  </parent>
 
-    <artifactId>jitsi-utils-kotlin</artifactId>
+  <artifactId>jitsi-utils-kotlin</artifactId>
+  <packaging>bundle</packaging>
 
-    <properties>
-        <kotlin.version>1.3.71</kotlin.version>
-    </properties>
+  <properties>
+    <kotlin.version>1.3.71</kotlin.version>
+  </properties>
 
-    <dependencies>
-        <dependency>
-            <groupId>org.jitsi</groupId>
-            <artifactId>jitsi-utils</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.jetbrains.kotlin</groupId>
-            <artifactId>kotlin-stdlib-jdk8</artifactId>
-            <version>${kotlin.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.jetbrains.kotlin</groupId>
-            <artifactId>kotlin-reflect</artifactId>
-            <version>${kotlin.version}</version>
-        </dependency>
+  <dependencies>
+    <dependency>
+      <groupId>org.jitsi</groupId>
+      <artifactId>jitsi-utils</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jetbrains.kotlin</groupId>
+      <artifactId>kotlin-osgi-bundle</artifactId>
+      <version>${kotlin.version}</version>
+    </dependency>
 
-        <dependency>
-            <groupId>org.jetbrains.kotlin</groupId>
-            <artifactId>kotlin-test</artifactId>
-            <version>${kotlin.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>io.kotlintest</groupId>
-            <artifactId>kotlintest-runner-junit5</artifactId>
-            <version>3.3.2</version>
-            <scope>test</scope>
-        </dependency>
-    </dependencies>
+    <dependency>
+      <groupId>org.jetbrains.kotlin</groupId>
+      <artifactId>kotlin-test</artifactId>
+      <version>${kotlin.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.kotlintest</groupId>
+      <artifactId>kotlintest-runner-junit5</artifactId>
+      <version>3.3.2</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
 
-    <build>
-        <sourceDirectory>src/main/kotlin</sourceDirectory>
-        <testSourceDirectory>src/test/kotlin</testSourceDirectory>
-        <plugins>
-            <plugin>
-                <groupId>org.jetbrains.kotlin</groupId>
-                <artifactId>kotlin-maven-plugin</artifactId>
-                <version>${kotlin.version}</version>
-                <executions>
-                    <execution>
-                        <id>compile</id>
-                        <phase>compile</phase>
-                        <goals>
-                            <goal>compile</goal>
-                        </goals>
-                    </execution>
-                    <execution>
-                        <id>test-compile</id>
-                        <phase>test-compile</phase>
-                        <goals>
-                            <goal>test-compile</goal>
-                        </goals>
-                    </execution>
-                </executions>
-                <configuration>
-                    <jvmTarget>1.8</jvmTarget>
-                </configuration>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <version>2.22.2</version>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-source-plugin</artifactId>
-                <version>3.1.0</version>
-                <executions>
-                    <execution>
-                        <id>attach-sources</id>
-                        <goals>
-                            <goal>jar</goal>
-                        </goals>
-                    </execution>
-                </executions>
-                <configuration>
-                </configuration>
-            </plugin>
-	    <plugin>
-		<groupId>org.apache.maven.plugins</groupId>
-		<artifactId>maven-antrun-plugin</artifactId>
-		<version>1.8</version>
-		<executions>
-		    <execution>
-			<id>ktlint</id>
-			<phase>verify</phase>
-			<configuration>
-			    <target name="ktlint">
-				<java taskname="ktlint" dir="${basedir}" fork="true" failonerror="true"
-				      classpathref="maven.plugin.classpath" classname="com.pinterest.ktlint.Main">
-				    <arg value="src/**/*.kt"/>
-				    <!-- to generate report in checkstyle format prepend following args: -->
-				    <!--
-					<arg value="[dash][dash]reporter=plain"/>
-					<arg value="[dash][dash]reporter=checkstyle,output=${project.build.directory}/ktlint.xml"/>
-					-->
-				    <!-- see https://github.com/shyiko/ktlint#usage for more -->
-				</java>
-			    </target>
-			</configuration>
-			<goals><goal>run</goal></goals>
-		    </execution>
-		    <execution>
-			<id>ktlint-format</id>
-			<configuration>
-			    <target name="ktlint">
-				<java taskname="ktlint" dir="${basedir}" fork="true" failonerror="true"
-				      classpathref="maven.plugin.classpath" classname="com.pinterest.ktlint.Main">
-				    <arg value="-F"/>
-				    <arg value="src/**/*.kt"/>
-				</java>
-			    </target>
-			</configuration>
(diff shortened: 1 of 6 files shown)


bbaldino:
What do the OSGi changes do?

ibauersachs:
They generate a manifest so the project can be used in a proper OSGi environment. There's a problem  though: the Kotlin and Java code share package names.
The maven-bundle-plugin now creates a single jar that contains the Java-classes in the -kotlin module. I actually think is fine, as Kotlin-based projects can just pull that one in, while Java-only projects stick with the regular version.

There are other, less elegant solutions:
- make jitsi-utils-kotlin an OSGi fragment
- move classes in jitsi-utils-kotlin to a different package

I converted this PR to a draft until I tried it in jitsi-rtp.

bbaldino:
Is this to load utils as an osgi bundle?  What's the use case for that?  Would it be possible to do the osgi-wrapping of these artifacts separately?

ibauersachs:
Yes. I'm updating everything used for Jitsi Desktop

> Is this to load utils as an osgi bundle?
Yes

> What's the use case for that?

Jitsi Desktop. I'm updating all dependencies and instead of backporting renewed/new stuff from media-transform and rtp, I'd like to update libjitsi. Not sure if that's possible, but I'll try anway.

> Would it be possible to do the osgi-wrapping of these artifacts separately?

What do you mean by separately? Keeping the jar content as it was?

bbaldino:
>> What's the use case for that?

> Jitsi Desktop. I'm updating all dependencies and instead of backporting renewed/new stuff from media-transform and rtp, I'd like to update libjitsi. Not sure if that's possible, but I'll try anway.

Sorry, I'm pretty OSGi-ignorant, but why would you want to dynamically load a utils lib as opposed to a normal dependency?

> What do you mean by separately? Keeping the jar content as it was?

Again, I don't know much about OSGi, but I'm wondering if we can have a plain artifact, and another artifact which 'wraps' the plain lib and bundles it for OSGi properly.

ibauersachs:
Having a jar OSGi ready simply means having an augmented MANIFEST.MF (declaring the required packages (import) and what it provides (export)). Having two separate versions of an artifact just for a slightly bigger text file is overkill.

bbaldino:
> Having a jar OSGi ready simply means having an augmented MANIFEST.MF (declaring the required packages (import) and what it provides (export)). Having two separate versions of an artifact just for a slightly bigger text file is overkill.

The issue is more that we don't use OSGi for this and few of us even understand what this is doing/how it works (as exhibited by my dumb questions here), which makes it more difficult for us to do other work in the pom/increases the risk of us breaking it.

ibauersachs:
> > Having a jar OSGi ready simply means having an augmented MANIFEST.MF (declaring the required packages (import) and what it provides (export)). Having two separate versions of an artifact just for a slightly bigger text file is overkill.
> 
> The issue is more that we don't use OSGi for this

`this`?

> and few of us even understand what this is doing/how it works (as exhibited by my dumb questions here), which makes it more difficult for us to do other work in the pom/increases the risk of us breaking it.

Well, OSGi is the foundation of Jitsi Desktop, which in turn is still used in Jigasi. Unless there are plans to rewrite Jigasi which I'm not aware of, I think it'd be better to keep some small POM changes to be able to use the same libraries everywhere.

What in particular are you afraid of breaking that can't be fixed?
In any case @damencho is still aware of all things OSGi related and available to you even when I'm not around.

bbaldino:
According to what I'm reading, a packaging type of bundle means we can't depend on it as a normal jar?  What would it look like for other projects to depend on this lib?

Also, is the purpose of this as a bundle for jitsi-desktop to be able to install something else in its place? Is that really a useful scenario?

ibauersachs:
Making a fragment is not possible, so the only options left are this PR or moving the Kotlin classes into a different package.

Btw, split packages are not just an OSGi issue, it will pop up with Java 9+ Modules as well.

ibauersachs:
> According to what I'm reading, a packaging type of bundle means we can't depend on it as a normal jar? What would it look like for other projects to depend on this lib?

Are you referring to something Kotlin specific? Otherwise, no, just use this dependency as usual.

> Also, is the purpose of this as a bundle for jitsi-desktop to be able to install something else in its place? Is that really a useful scenario?

No, not to install something else, but to install _this_ bundle.

bbaldino:
> Making a fragment is not possible

What is this referring to?

> Are you referring to something Kotlin specific? Otherwise, no, just use this dependency as usual.

Ok cool, what I read must have be wrong or out-of-date.

> No, not to install something else, but to install this bundle.

I think that's what I meant, the purpose of an OSGi bundle at all is to make the actual instance of a particular bundle which is loaded flexible at runtime, right?  That doesn't seem very useful for something like utils (or maybe I've misunderstood).

bbaldino:
Also, there is really no reason for these 2 chunks of code to be in separate artifacts at all: a pure java library would be able to depend on it just fine.  I did the new library to smooth the transition into some kotlin stuff.  Maybe it's better if we just merged them.

ibauersachs:
> > Making a fragment is not possible
> 
> What is this referring to?

One of the alternatives I mentioned before, but I was wrong there.

> > Are you referring to something Kotlin specific? Otherwise, no, just use this dependency as usual.
> 
> Ok cool, what I read must have be wrong or out-of-date.
> 
> > No, not to install something else, but to install this bundle.
> 
> I think that's what I meant, the purpose of an OSGi bundle at all is to make the actual instance of a particular bundle which is loaded flexible at runtime, right? That doesn't seem very useful for something like utils (or maybe I've misunderstood).

Yes, but if it isn't a bundle, it can't load it at all.

bbaldino:
> Yes, but if it isn't a bundle, it can't load it at all.

Can it not declare a normal dependency?

ibauersachs:
> Also, there is really no reason for these 2 chunks of code to be in separate artifacts at all: a pure java library would be able to depend on it just fine. I did the new library to smooth the transition into some kotlin stuff. Maybe it's better if we just merged them.

I'm fine with that if you are all fine with the requirement to include the Kotlin stdlib everywhere...

ibauersachs:
> > Yes, but if it isn't a bundle, it can't load it at all.
> 
> Can it not declare a normal dependency?

There are two options, but all of them ugly:
- Provide it on the system classpath and list all contained packages manually
- Let some script at build time figure out the possible imports and fiddle them into a wrapped jar, which is likely incorrect and would loose the Maven dependency tree
- Include the util.jar into the dependee (as a resource) and put it onto the dependee's classpath (note that in OSGi, every bundle has its own classloader! we'd end up with multiple copies in memory)

bbaldino:
> > Also, there is really no reason for these 2 chunks of code to be in separate artifacts at all: a pure java library would be able to depend on it just fine. I did the new library to smooth the transition into some kotlin stuff. Maybe it's better if we just merged them.
> 
> I'm fine with that if you are all fine with the requirement to include the Kotlin stdlib everywhere...

Pretty sure there's no need to explicitly require this, it would be pulled in automatically as a transitive dependency.

bbaldino:
> > > Yes, but if it isn't a bundle, it can't load it at all.
> > 
> > 
> > Can it not declare a normal dependency?
> 
> There are 3 options, but all of them ugly:
> 
> * Provide it on the system classpath and list all contained packages manually
> * Let some script at build time figure out the possible imports and fiddle them into a wrapped jar, which is likely incorrect and would loose the Maven dependency tree
> * Include the util.jar into the dependee (as a resource) and put it onto the dependee's classpath (note that in OSGi, every bundle has its own classloader! we'd end up with multiple copies in memory)

Ok, I assumed jitsi-desktop was using maven and could just list the dependency, but I guess that isn't the case.

ibauersachs:
> Pretty sure there's no need to explicitly require this, it would be pulled in automatically as a transitive dependency.

Yes, it will. This was more meant for political consideration rather than technical.

bbaldino:
> > Pretty sure there's no need to explicitly require this, it would be pulled in automatically as a transitive dependency.
> 
> Yes, it will. This was more meant for political consideration rather than technical.

Btw, jitsi-desktop is already pulling in jitsi-utils, so the purpose of this is to be able to use code in jitsi-utils-kotlin in jitsi-desktop as well?  So won't this already be the case after this PR?

ibauersachs:
> Ok, I assumed jitsi-desktop was using maven and could just list the dependency, but I guess that isn't the case.

So far it's an Ant build and some POMs for the Meet ecosystem. I'm currently reorganizing/rewriting this to Gradle. In any case, the build system is irrelevant for OSGi unless you use a bnd-workspace.

bbaldino:
> > Ok, I assumed jitsi-desktop was using maven and could just list the dependency, but I guess that isn't the case.
> 
> So far it's an Ant build and some POMs for the Meet ecosystem. I'm currently reorganizing/rewriting this to Gradle. In any case, the build system is irrelevant for OSGi unless you use a bnd-workspace.

Ok, but my question was about whether or not an explicit dependency could be declared (not only a bundle-style dependency)

ibauersachs:
> Btw, jitsi-desktop is already pulling in jitsi-utils, so the purpose of this is to be able to use code in jitsi-utils-kotlin in jitsi-desktop as well, right?

Not directly. My current goal is to update libjitsi with jitsi-media-transform and jitsi-rtp. Not sure if this is possible, but without the OSGi support here I can't even try.

> So won't this already be the case after this PR?

jitsi-utils-kotlin/pom.xml
```diff
@@ -1,148 +1,136 @@
...
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.jetbrains</groupId>
+      <artifactId>annotations</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.jetbrains.kotlin</groupId>
+      <artifactId>kotlin-osgi-bundle</artifactId>
```
bbaldino, line 45: So I understand: why does this obsolesce the requirement for kotlin-reflect?

ibauersachs: > To enable Kotlin OSGi support you need to include kotlin-osgi-bundle instead of regular Kotlin libraries. It is recommended to remove kotlin-runtime, kotlin-stdlib and kotlin-reflect dependencies as kotlin-osgi-bundle already contains all of them.

https://kotlinlang.org/docs/reference/kotlin-osgi.html

bbaldino:
> > Btw, jitsi-desktop is already pulling in jitsi-utils, so the purpose of this is to be able to use code in jitsi-utils-kotlin in jitsi-desktop as well, right?
> 
> Not directly. My current goal is to update libjitsi with jitsi-media-transform and jitsi-rtp. Not sure if this is possible, but without the OSGi support here I can't even try.

Why's that?  libjitsi can express a direct dependency on this lib without using a bundle.

ibauersachs:
> Why's that? libjitsi can express a direct dependency on this lib without using a bundle.

It compiles, but you can't use it.

jitsi-utils-kotlin/pom.xml
bbaldino: Hm, also from there:

> Most regular Kotlin dependencies are not OSGi-ready, so you shouldn't use them and should remove them from your project.

This is the sort of thing that doesn't make me thrilled about changing this to be OSGi.

bbaldino:
> > Why's that? libjitsi can express a direct dependency on this lib without using a bundle.
> 
> It compiles, but you can't use it.

I don't follow...libjitsi has all sorts of "regular" dependencies?

jitsi-utils-kotlin/pom.xml
ibauersachs: The non-readyness is true for many, if not most, regular Java libraries as well. It doesn't mean it's not possible to use them, but it will require more effort in applications running in an OSGi environment, i.e. Jitsi Desktop. All other applications won't bother.

ibauersachs:
> > > Why's that? libjitsi can express a direct dependency on this lib without using a bundle.
> > 
> > 
> > It compiles, but you can't use it.
> 
> I don't follow...libjitsi has all sorts of "regular" dependencies?

No it doesn't, they are all OSGi ready.

bbaldino:
> > > > Why's that? libjitsi can express a direct dependency on this lib without using a bundle.
> > > 
> > > 
> > > It compiles, but you can't use it.
> > 
> > 
> > I don't follow...libjitsi has all sorts of "regular" dependencies?
> 
> No it doesn't, they are all OSGi ready.

Ah, I see what you mean.  Still, if the goal is to move libjitsi over, it would probably be best to have a test harness in libjitsi itself to verify these sorts of things rather than testing it within jitsi-desktop which would at least delay the need for bundling here until you know porting over libjitsi is possible.

jitsi-utils-kotlin/pom.xml
bbaldino: So, another dumb question, is the 'bundle-able' requirement for all dependencies?  That is, must all the libs `jitsi-utils-kotlin` depend be bundle-friendly?  Or is it just the direct dependency itself (jitsi-utils-kotlin)?

ibauersachs:
> Still, if the goal is to move libjitsi over, it would probably be best to have a test harness in libjitsi itself to verify these sorts of things rather than testing it within jitsi-desktop which would at least delay the need for bundling here until you know porting over libjitsi is possible.

Of course I can leave it be for the time being. But when will jitsi-srtp require something from jitsi-utils-kotlin? I'd agree with waiting if it were something that causes harm, but it doesn't.

jitsi-utils-kotlin/pom.xml
ibauersachs: It's the entire chain. In jitsi-rtp e.g. the `unsigned` library is such a non-osgi dependency that would hinder things. But since Kotlin itself now has unsigned types (although experimental, but other experimental features are also enabled there)...

bbaldino: So, this would require everything `jitsi-utils-kotlin` depends on to be bundle-friendly?

ibauersachs: Yes, this is what JetBrains means with `Most regular Kotlin dependencies ...`

bbaldino: > It's the entire chain. In jitsi-rtp e.g. the `unsigned` library is such a non-osgi dependency that would hinder things. But since Kotlin itself now has unsigned types (although experimental, but other experimental features are also enabled there)...

But didn't you update libjitsi to use jitsi-srtp?  Is it just that you haven't changed jitsi-desktop to use that version of libjitsi yet?

ibauersachs: Check your e-mail for a Meet-link please :-)

PR #60, opened 2020-05-24
Build improvements

ibauersachs:
Closes #48
Closes #59


jitsi-utils/pom.xml
```diff
@@ -51,66 +43,52 @@
...
-      <version>4.11</version>
+      <version>4.13</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
-
+  
   <build>
     <plugins>
-      <plugin>
```
JonathanLennox, line 79: Did you mean to drop the ktlint?

ibauersachs: Not sure I follow, this comment is on the java package and ktlint is in jitsi-utils-kotlin:
https://github.com/jitsi/jitsi-utils/pull/60/files/4a6d13e7698c49458ac89d5f5e12ae374bbc9d3a..83bb7396074e495693957870f89d0b3fb8fd44aa#diff-a11cc5b682899cc8f742fc654588e716R78

JonathanLennox: Oh, I see, you changed from 'antrun ktlint' to 'ktlint-maven-plugin'.  Is the behavior the same?

I don't know why GitHub reported this comment as being on `jitsi-utils/pom.xml`, I commented on `jitsi-utils-kotlin/pom.xml`

ibauersachs: I didn't change any Kotlin code, but I ran the plugin to format code and it didn't change, i.e. reformat, anything. The maven plugin uses the same libraries that were invoked from antrun.

.editorconfig
```diff
@@ -0,0 +1,3 @@
+[pom.xml]
+indent_style = space
+indent_size = 2
```
bbaldino, line 3: I find XML hard enough to parse without having such a small indentation difference.  Can we leave out the indent space?  I have a feeling it'll just cause debate (see: this very comment :P)

jitsi-utils/src/main/java/org/jitsi/utils/PasswordUtil.java
```diff
@@ -71,7 +71,7 @@ public static String replacePasswords(String string, String[] passwordArgs)
     {
         for (String passArg : passwordArgs)
         {
-            if (StringUtils.isNullOrEmpty(passArg))
+            if (passArg == null || passArg.trim().isEmpty())
```
bbaldino, line 74: I think we'd started to move _towards_ apache commons for these (and it doesn't look like it got removed), is there a reason to swap to the manual check here?

.editorconfig
ibauersachs: 4 spaces? I don't care either way, I just want it to be _consistent_.

jitsi-utils/src/main/java/org/jitsi/utils/PasswordUtil.java
ibauersachs: jitsi-utils doesn't have a dependency on apache commons and I didn't want to pull it in for this _one_ usage.

PR #63, opened 2020-06-17
Removes the deprecation of MediaType.DATA.

bgrozev:


ibauersachs:
That deprecation was there for a reason... what's the reason to intentionally continue to violate RFC 4566?

bbaldino:
I can't speak for boris and don't know the whole history, but in practice the usage of that enum has outgrown a specific mapping from RFC 4566; `DATA` is still used in Colibri and the deprecation was annoying/misleading (since that value is not deprecated in Colibri).  Perhaps the comment relating it to 4566 should be removed/re-worded, though.

ibauersachs:
I made that deprecation and the idea was to move away from it. If Colibri is still using it, Colibri should be fixed, not a in fact not misleading warning removed.

PR #65, opened 2020-07-14
ref: Use a generic ID instead of SSRC (long) to index speakers.

bgrozev:


bbaldino:
Do we use different types for this?

bgrozev:
> Do we use different types for this?

I think I'm missing something. Was the comment meant for a specific line of code?

Currently we index using the stream's SSRC, which leads the endpoints having to be looked up by SSRC. The change in the bridge uses the endpoint IDs.

bbaldino:
> > Do we use different types for this?
> 
> I think I'm missing something. Was the comment meant for a specific line of code?
> 
> Currently we index using the stream's SSRC, which leads the endpoints having to be looked up by SSRC. The change in the bridge uses the endpoint IDs.

Yeah, sorry, I was referring to using an `Object` here...so is it a temporary thing to enable the transition as the bridge moves from SSRC to ep ID?  Or just to change it such that this code doesn't care what's used?

bgrozev:
I wanted to make it generic. Hold on, I'll actually make it generic and not an `Object`.

PR #66, opened 2020-07-30
Add SafeExecutor and SafeScheduledExecutor.

bgrozev:


jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
```diff
@@ -0,0 +1,120 @@
...
+}
+
+class ExecutorShutdownTimeoutException : Exception("Timed out trying to shutdown executor service")
+
+class SafeExecutor(
+    name: String,
+    delegate: ExecutorService
+) : TaskServiceWrapper<ExecutorService>(name, delegate) {
+
+    fun execute(task: Runnable) = delegate.execute(wrapTask(task))
```
bbaldino, line 84: It looks like you mimic the `Executor` interface here, but also have `ExecutorService`-like methods (the submit-style methods).  Why not just mimic the `ExecutorService` API (which is what we use everywhere) for both types and have the `submit` calls and `unsafeSubmit` variants?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
```diff
@@ -0,0 +1,120 @@
...
+    fun <T> unsafeSubmit(task: Runnable, result: T): Future<T>? = delegate.submit(task, result)
+    fun unsafeSubmit(task: Runnable): Future<*>? = delegate.submit(task)
+}
+
+class SafeScheduledExecutor(
+    name: String,
+    delegate: ScheduledExecutorService
+) : TaskServiceWrapper<ScheduledExecutorService>(name, delegate) {
+
+    fun schedule(command: Runnable, delay: Long, unit: TimeUnit) {
```
bbaldino, line 97: Do we need a `KFunction0<Unit>` version of this?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
```diff
@@ -0,0 +1,120 @@
...
+        unsafeSchedule(wrapTask(command), delay, unit)
+    }
+
+    fun unsafeSchedule(command: Runnable, delay: Long, unit: TimeUnit): ScheduledFuture<*> =
+        delegate.schedule(command, delay, unit)
+
+    fun unsafeSchedule(command: KFunction0<Unit>, delay: Duration): ScheduledFuture<*> =
+        unsafeSchedule(Runnable { command() }, delay.toMillis(), TimeUnit.MILLISECONDS)
+
+    fun unsafeScheduleAtFixedRate(
```
bbaldino, line 107: What about safe versions of `scheduleAtFixedRate` and `scheduleWithFixedDelay` ?

jitsi-utils/src/main/java/org/jitsi/utils/queue/AsyncQueueHandler.java
```diff
@@ -45,16 +45,23 @@
      * The default <tt>ExecutorService</tt> to run <tt>AsyncQueueHandler</tt>
      * when there is no executor injected in constructor.
      */
-    private final static ExecutorService sharedExecutor
-        = Executors.newCachedThreadPool(
-            new CustomizableThreadFactory(
-                AsyncQueueHandler.class.getName() + "-executor-", true));
+    private final static Executor sharedExecutor = new Executor() {
```
bbaldino, line 48: So this is not using the safe variant?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
bgrozev: `ExecutorService` extends `Executor` (so it inherits `void execute`), which is why I named the "safe" method `execute`.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
bgrozev: Sure

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
bgrozev: I can add all of these, thought I don't know how useful they are.

jitsi-utils/src/main/java/org/jitsi/utils/queue/AsyncQueueHandler.java
bgrozev: No, it uses the `Future`. It does it's own exception handling, at [least for the `handleItem` method](https://github.com/jitsi/jitsi-utils/blob/master/jitsi-utils/src/main/java/org/jitsi/utils/queue/AsyncQueueHandler.java#L139)

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
bbaldino: Sorry, I don't think I followed this.  Why expose `execute` at all, given nothing we use uses `Executor` directly, but instead `ExecutorService`/`ScheduledExecutorService`?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
bbaldino: It looks like we use `scheduleAtFixedRate` in `Conference.java` in the bridge?

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
bgrozev: To mimic the `ExecutorService` API -- the method which runs a task and doesn't return anything is named `execute`, while `submit` methods return a future.

bbaldino: Ok, from what I saw we didn't use `execute` anywhere in the code, so think I was confused.  Are you thinking that we should change the places which don't use the future over to `execute` (from `submit`)? That's probably a good idea.

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
bgrozev: Yes, that's now `unsafeScheduleAtFixedRate`

jitsi-utils-kotlin/src/main/kotlin/org/jitsi/utils/concurrent/Executors.kt
bgrozev: Yes, that's what I did in the PRs in jmt/jvb

PR #69, opened 2020-08-04
merge jitsi-utils-kotlin and jitsi-utils

bbaldino:


bbaldino:
@ibauersachs wrote you an email about this, but not sure if it was the right address.

bbaldino:
@ibauersachs I found that kotlin has an OSGi-ready version, so I used that instead...hopefully that should work fine then.

ibauersachs:
The address is correct, thanks for the heads-up. I think the changes are okay and if something is missing I'll deal with a follow-up PR.
But is there a reason to keep the Maven submodule? I think it just complicates the build files.

bbaldino:
> But is there a reason to keep the Maven submodule? I think it just complicates the build files.

We've been thinking lately about more, smaller artifacts in a repo rather than a single big one, so I left it thinking we may want to take advantage of that in the future.  Normally I'd say get rid of it and we can add it back, but in the past it's been such a pain to migrate to submodules my thought was to leave it.

jitsi-utils/pom.xml
```diff
@@ -127,32 +120,90 @@
...
+                        <id>java-test-compile</id>
                         <phase>test-compile</phase>
                         <goals>
                             <goal>testCompile</goal>
                         </goals>
                     </execution>
                 </executions>
             </plugin>
+            <!-- This plugin is necessary for free/extensions functions defined in the kotlin code to be
+            visible to kotlin code using this lib.  See https://youtrack.jetbrains.com/issue/KT-9164. -->
```
ibauersachs, line 186: I assume you tried the other method mentioned in this issue?
```
<Include-Resource>
     /=src/main/resources/,
     META-INF=target/classes/META-INF
 </Include-Resource>
```

ibauersachs: (That would go into <instructions> below line 141)

ibauersachs:
> > But is there a reason to keep the Maven submodule? I think it just complicates the build files.
> 
> We've been thinking lately about more, smaller artifacts in a repo rather than a single big one, so I left it thinking we may want to take advantage of that in the future. Normally I'd say get rid of it and we can add it back, but in the past it's been such a pain to migrate to submodules my thought was to leave it.

Keeping things in advance is not something I like too much, but if you think this will really in this repo, keep it.

jitsi-utils/pom.xml
bbaldino: Cool, looks like this worked, I'll do it this way.

bbaldino:
> > > But is there a reason to keep the Maven submodule? I think it just complicates the build files.
> > 
> > 
> > We've been thinking lately about more, smaller artifacts in a repo rather than a single big one, so I left it thinking we may want to take advantage of that in the future. Normally I'd say get rid of it and we can add it back, but in the past it's been such a pain to migrate to submodules my thought was to leave it.
> 
> Keeping things in advance is not something I like too much, but if you think this will really happen in this repo, keep it.

Ok, I ended up getting rid of it.

src/main/kotlin/org/jitsi/utils/logging2/Foo.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jitsi.utils.logging2
+
+class Foo {
```
ibauersachs, line 19: Uhh...?

pom.xml
```diff
@@ -60,15 +199,6 @@
                     </compilerArgs>
                 </configuration>
             </plugin>
-            <plugin>
```
ibauersachs, line 63: Is this causing issues with Kotlin?

src/main/kotlin/org/jitsi/utils/logging2/Foo.kt
bbaldino: Oops.

pom.xml
bbaldino: Nope, this removal was unintentional.  Added it back.

pom.xml
```diff
@@ -17,41 +17,180 @@
...
+                            META-INF=target/classes/META-INF
+                        </Include-Resource>
+                    </instructions>
+                </configuration>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.8.1</version>
+                <version>3.5.1</version>
```
bgrozev, line 167: Change from 3.8.1 to 3.5.1?

bbaldino: Nice catch, must have been this version in the submodule file. Fixed.

PR #71, opened 2020-09-21
feat: Add RateTracker with support for bucketSize > 1.

bgrozev:


src/main/kotlin/org/jitsi/utils/stats/RateTracker.kt
```diff
@@ -0,0 +1,156 @@
...
+
+open class RateTracker @JvmOverloads constructor(
+    /**
+     * The number of buckets.
+     */
+    private val numBuckets: Int,
+    /**
+     * The size of each bucket in milliseconds.
+     */
+    private val bucketSizeMs: Int = 1,
```
bbaldino, line 31: Is this in ms (as opposed to a `Duration`) for backwards compatibility?  Or perf?

src/main/kotlin/org/jitsi/utils/stats/RateTracker.kt
```diff
@@ -0,0 +1,156 @@
...
+* limitations under the License.
+*/
+package org.jitsi.utils.stats
+
+import org.jitsi.utils.ms
+import java.lang.IllegalArgumentException
+import java.time.Clock
+import java.time.Duration
+
+open class RateTracker @JvmOverloads constructor(
```
bbaldino, line 23: Docs

src/main/kotlin/org/jitsi/utils/stats/RateTracker.kt
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Bucket index of oldest counter recorded in buckets.
+     */
+    private var oldestIndex = 0
+
+    /**
+     * Oldest time recorded in buckets.
+     */
+    private var oldestTime: Long = 0
```
bbaldino, line 72: This is a timestamp in `ms`?

src/main/kotlin/org/jitsi/utils/stats/RateTracker.kt
```diff
@@ -0,0 +1,156 @@
...
+     */
+    val windowSizeMs = numBuckets * bucketSizeMs
+
+    /**
+     * Convert a clock time to the local time representation (one tick is [bucketSizeMs] milliseconds).
+     */
+    private fun coerceMs(timeMs: Long) = timeMs / bucketSizeMs
+
+    @Synchronized
+    private fun eraseOld(now: Long) {
```
bbaldino, line 85: Is now a timestamp in ms?

src/main/kotlin/org/jitsi/utils/stats/RateTracker.kt
```diff
@@ -0,0 +1,156 @@
...
+        val newOldestTime = now - buckets.size + 1
```
bbaldino, line 86: I'll probably have to go over all of this some more, but this is confusing to read: we subtract the number of buckets from a timestamp and get a timestamp?

bgrozev: Yes, because the timestamp is in "ticks per bucketSizeMs". Note this class is a translation of `RateStatistics` with this feature (bucket size > 1) added on top.

src/main/kotlin/org/jitsi/utils/stats/RateTracker.kt
bgrozev: No, that's why I renamed it from `nowMs` to `now`. I'll add a note.

src/main/kotlin/org/jitsi/utils/stats/RateTracker.kt
bgrozev: No, it's ms converted to "one tick per bucketSizeMs".

src/main/kotlin/org/jitsi/utils/stats/RateTracker.kt
bgrozev: 👍

src/main/kotlin/org/jitsi/utils/stats/RateTracker.kt
bgrozev: Neither. I'm only keeping `RateStatistics` for backward comp. I'll clean that up a bit.

bgrozev:
> Dumb question, but what does a bucket imply here? If the bucket size is 5ms, does that mean that all values passed to `update` within a 5ms span are merged to a single value?

Yes, exactly. The reason I'm adding this is because I want to keep track of packet loss for the last 1 minute, and using 1ms resolution is not necessary and requires a lot of memory: `60 seconds * 1000 ms/sec * 8 bytes in a long` = 480KB

JonathanLennox:
The old `RateStatistics`'s `scale` parameter had the feature that allowed it to do automatic bytes-to-bits conversion (i.e. pass in bytes, get out bits per second).  Do you want to enable this in `RateTracker`, rather than just in the wrapper?

bgrozev:
> The old `RateStatistics`'s `scale` parameter had the feature that allowed it to do automatic bytes-to-bits conversion (i.e. pass in bytes, get out bits per second). Do you want to enable this in `RateTracker`, rather than just in the wrapper?

I kept the wrapper just for backward compat. For bits/bytes I was thinkingn we'd use `BitrateTracker`, which had to be in JMT.

JonathanLennox (approved):

PR #75, opened 2020-11-18
Deploy to Maven Central

ibauersachs:

PR #76, opened 2020-11-19
Fix release with Dokka

ibauersachs:
Follow-up to #75. Javadoc does not work on Java 11 with Kotlin, instead use Dokka to also publish Kotlin documention.

Note that the `<pluginRepositories>` tag in the pom is really annoying, this shouldn't be in a pom published to Maven Central. Remove it once kotlin/dokka#41 is fixed. Plus, connections to JCenter/Bintray are unstable, which might cause shaky builds (but only if the pom changed, otherwise the cache kicks in).

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -34,7 +34,7 @@ jobs:
             ${{ runner.os }}-maven-
 
       - name: Build and test with Maven
-        run: mvn -B verify -Drelease=true
+        run: mvn -B verify
 
   release:
     if: github.ref == 'refs/heads/master'
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -54,6 +54,14 @@ jobs:
           server-username: SONATYPE_USER
           server-password: SONATYPE_PW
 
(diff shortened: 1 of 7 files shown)


damencho (approved):

ibauersachs:
@damencho @bbaldino  For some reason it does not show yet at [search.maven.org](https://search.maven.org/search?q=g:org.jitsi), but really is deployed: https://repo1.maven.org/maven2/org/jitsi/jitsi-utils/ 🎉

damencho:
Nice! Thank you 🎉  Yeah I was checking in the morning and was not seeing it. Maybe it will take time for the web part ... but if it is deployed it can be used, which is the important part!

bbaldino:
We didn't lose the source jar here I hope/assume, yeah?  Is something else doing that now that the `maven-source-plugin` is removed?

ibauersachs:
@bbaldino Good catch! That's a mistake, I'll fix that later tonight. The idea was that -DperformRelease creates/attaches the sources automatically, but since I had to get rid of this because of the Javadoc vs. Dokka woes, it's no longer there. The snippet can be restored as is.

PR #80, opened 2020-12-02
Allow manipulation of logger delegate handlers

bbaldino:
Jibri doesn't currently use our logger because it needs to override the logging handlers so it can log to different files, so this PR expose  some java.util.logging APIs to be enable that.

### DIFF
--- a/src/test/java/org/jitsi/utils/logging2/LoggerImplTest.java
+++ b/src/test/java/org/jitsi/utils/logging2/LoggerImplTest.java
@@ -186,6 +186,40 @@ public class LoggerImplTest
         logger.info(() -> "hello, world!");
         assertEquals(1, fakeLogger.logLines.size());
     }
+
+    @Test
+    public void testHandler()
+    {
+        // We want to use a real logger for this test
+        LoggerImpl.loggerFactory = oldLoggerFactoryFunction;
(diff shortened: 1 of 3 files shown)


src/test/java/org/jitsi/utils/logging2/LoggerImplTest.java
```diff
@@ -186,6 +186,40 @@ public void testMsgSupplier()
...
+        FakeHandler handler = new FakeHandler();
+        logger.useHandler(handler);
+
+        logger.info("hello, world!");
+
+        assertEquals(1, handler.logRecords.size());
+        LogRecord record = handler.logRecords.get(0);
+        assertEquals(record.getMessage(), "hello, world!");
+        assertEquals(record.getLoggerName(), "test");
+    }
```
JonathanLennox, line 207: Is there an easy way to test that the old parent handler didn't get used?

bbaldino: Hm, not that I know of.  But since `setUseParentHandlers` is a direct pass through to the java.util.Logging method there's not too much room for things to go wrong.

JonathanLennox (approved):

PR #82, opened 2020-12-16
Include Kotlin code in generated -sources.jar

ibauersachs:


pom.xml
```diff
@@ -244,6 +243,24 @@
...
+                <version>3.2.0</version>
+                <executions>
+                    <execution>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>add-source</goal>
+                        </goals>
+                        <configuration>
+                            <sources>
+                                <source>src/main/kotlin</source>
```
bbaldino, line 258: Do we need `src/main/java` here as well? What's grabbing those sources?

ibauersachs: No, the Java sources are taken by default. maven-sources-plugin consumes what is known in the Maven Reactor as sources, and unfortunately the kotlin-maven-plugin doesn't mark src/main/kotlin as sources.

bbaldino: Ok, and this setting won't override the existing one, right?  I.e. now only do the kotlin sources?

ibauersachs: No, of course not.

bbaldino (approved):

bbaldino:
I noticed the other day how long the docs/dokka task takes, wondering if we should only do that when a specific profile is used so when building locally we don't run it but we can do it for releases.

ibauersachs:
I'd prefer that too, but it has nothing to do with this PR.

bbaldino:
> I'd prefer that too, but it has nothing to do with this PR.

I know, this was just the easiest way to communicate it at the time :)

PR #84, opened 2021-01-20
feat: Add AsyncEventEmitter, use explicit names for fireEventSync vs Async.

bgrozev:


src/main/kotlin/org/jitsi/utils/event/EventEmitter.kt
```diff
@@ -33,3 +33,20 @@ class EventEmitter<EventHandlerType> {
...
+
+open class SyncEventEmitter<EventHandlerType> : EventEmitter<EventHandlerType>() {
+    fun fireEventSync(event: EventHandlerType.() -> Unit) {
+        eventHandlers.forEach { it.apply(event) }
+    }
+}
+
+class AsyncEventEmitter<EventHandlerType>(
+    private val executor: ExecutorService
+) : SyncEventEmitter<EventHandlerType>() {
```
bbaldino, line 45: What do you think about making `SyncEventEmitter` and `AsyncEventEmitter` peers, each extending `EventEmitter`?  I'm torn on whether or not a `SyncEventEmitter` exposing sync-only, but `AsyncEventEmitter` exposing both sync and async APIs feels weird.  Sure `AsyncEventEmitter` _can_ function as both, but I go back and forth on whether or not "AsyncEventEmitter is-a SyncEventEmitter" feels right.

bgrozev: I think with explicit names it is be hard to make a mistake and unintentionally call `fireEventSync` on an `AsyncEventEmitter`, and if for some reason you want to use both, you'd be left with having to have two emitters.

bgrozev: What if there is no `SyncEventEmitter` and `fireEventSync` is part of `EventEmitter`?

bbaldino (approved):

PR #85, opened 2021-01-21
chore: Move nexus to a profile (allow deploy to non-nexus).

bgrozev:


bbaldino (approved):

ibauersachs:
@bgrozev Are you deploying this on some 8x8 internal infrastructure so that non-Nexus deployments are required? Or what other reason is there?

ibauersachs:
@bgrozev And you broke the deployment with the missing `\` at the EOL to continue the bash command.

bgrozev:
> @bgrozev And you broke the deployment with the missing `\` at the EOL to continue the bash command.

Fixed in 86 🤦

bgrozev:
We're pushing to jitsi-maven-repository on github in parallel, because the deployment to mavencentral keeps failing randomly.

bgrozev:
See e.g. https://github.com/jitsi/jitsi-utils/runs/1744100637?check_suite_focus=true

ibauersachs:
You realize you could have just clicked on `re-run jobs` there? Intermittent failures _can_ happen, even on `jitsi-maven-repository`. I don't think it's a good idea to have different binaries with the same version in multiple repos (different because they're not built reproducible, different Maven and JDK versions, etc.).

bgrozev:
> You realize you could have just clicked on `re-run jobs` there?

I had to re-run jicoco 4-5 times last week (over a period of a day or so) until it succeeded. Then it took over an hour for the artefact to actually show up in the repo. Then jitsi-utils also failed a week later.

ibauersachs:
Okay. Sonatype seems to have quite some issues during the past week(s): https://status.maven.org/

I have never seen it this bad.

PR #86, opened 2021-01-21
fix: Fix the gh action syntax.

bgrozev:


ibauersachs (approved):
I'm not sure the whole thing with -P works due to Maven's unorthodox handling of profile activation, but let's try.

PR #87, opened 2021-01-22
chore: Another attempt to fix deployment to mavencentral.

bgrozev:


ibauersachs (changes requested):
I think it would make more sense to move the nexus-staging-maven-plugin into the release profile (which is activated with the property) and remove the -P flag.

ibauersachs (approved):

bgrozev:
Okay, we're back to a failure `trying to close staging repository`.

ibauersachs:
Yes, OSSRH has massive issues. Not much we can do atm. See https://status.maven.org/

ibauersachs:
Up again and deployed.

PR #92, opened 2021-03-05
Generate code coverage reports with mvn verify

gpolitis:
This will create a code coverage report in target/site/jacoco every time the unit tests are executed.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -5,6 +5,10 @@ name: Java CI with Maven
 
 on: [push, pull_request]
 
+env:
+  # Java version to use for the release
+  RELEASE_JAVA_VERSION: 8
+
 jobs:
   test:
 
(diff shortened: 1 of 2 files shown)


bbaldino:
How much time (if any?) does this add to the build? I run `mvn verify` locally pretty often, and wondering if we should tie this to a profile that is only used on CI so we don't have to run it locally when we just want to run tests.

gpolitis:
> How much time (if any?) does this add to the build? I run `mvn verify` locally pretty often, and wondering if we should tie this to a profile that is only used on CI so we don't have to run it locally when we just want to run tests.

without JaCoCo

`mvn clean test  78.02s user 4.93s system 350% cpu 23.694 total`

with JaCoCo

`mvn clean test  94.47s user 4.60s system 393% cpu 25.188 total`

It adds a bit of delay apparently, it's probably a good idea to tie this to a profile that only runs on CI.

bbaldino (approved):

bbaldino (approved):

PR #97, opened 2021-04-20
Add stats to PacketQueue, and restructure how its stat listener works.

JonathanLennox:


src/main/java/org/jitsi/utils/queue/PacketQueue.java
```diff
@@ -129,8 +131,8 @@ public PacketQueue(
         {
             enableStatistics = enableStatisticsDefault;
         }
-        queueStatistics
-            = enableStatistics ? new QueueStatistics() : null;
+        observer
```
bbaldino: Maybe take in an optional `Observer` instead of `enableStatistics`?  It seems like this bit should be external, but maybe I'm not fully following.

bbaldino: This also feels a bit weird given there's a `setObserver` call as well, seems like it should just be an external class creating the observer and adding/setting it?

JonathanLennox: That's what the old (last year's) PR did - but it resulted in a bunch of duplicated code for all the observers, which seemed unnecessary, so I thought adding the shortcut for this normal case was useful.

bbaldino: So a caller can either set its own observer which it would access via the observer directly, or tell `PacketQueue` to enable statistics, where it creates its own observer and exposes stats for?  Seems like we'd only need one of those use cases.

src/main/kotlin/org/jitsi/utils/stats/BucketStats.kt
```diff
@@ -0,0 +1,137 @@
...
+                if (it.first < p999 && sum > 0.999 * totalCount) p999 = it.first
+            }
+
+            // Not enough data
+            if (totalCount < 100 || p99 == Long.MAX_VALUE) p99 = -1
+            if (totalCount < 1000 || p999 == Long.MAX_VALUE) p999 = -1
+            return Snapshot(bucketCounts, p99, p999)
+        }
+
+    private fun findBucket(value: Long): LongAdder {
```
bgrozev, line 107: Would you care to make this `protected open` in case anyone wants to optimize for a different distribution (just based on the comment below)?

src/main/kotlin/org/jitsi/utils/stats/BucketStats.kt
```diff
@@ -0,0 +1,137 @@
...
+
+package org.jitsi.utils.stats
+
+import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.LongAdder
+import org.jitsi.utils.OrderedJsonObject
+import org.jitsi.utils.maxAssign
+import java.lang.IllegalArgumentException
+
+open class BucketStats(thresholdsNoMax: LongArray, val averageMaxLabel: String, val bucketLabel: String) {
```
bgrozev, line 25: Any reason not to have default values (""?) for `aveageMaxLabel` and `bucketLabel`?

src/test/java/org/jitsi/utils/queue/PacketQueueBenchmarkTests.java
```diff
@@ -20,14 +20,15 @@
...
+import org.junit.jupiter.api.condition.*;
 
 /**
  * Contains test for checking performance aspects of various
  * PacketQueue configuration
  *
  * @author Yura Yaroshevich
  */
-@Disabled("Check only performance aspect of PacketQueue")
+@EnabledIfSystemProperty(named="org.jitsi.utils.doPerf", matches=".*")
```
bgrozev, line 31: Oh, this is neat!

src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
```diff
@@ -0,0 +1,225 @@
...
+        var TRACK_TIMES = false
+
+        private val queueStatsById = ConcurrentHashMap<String, QueueStatistics>()
+
+        internal fun globalStatsFor(queue: PacketQueue<*>, clock: Clock) = queueStatsById.computeIfAbsent(queue.id()) {
+            /* Assume all queues with the same ID have the same capacity and can use the same size buckets. */
+            QueueStatistics(queue.capacity(), clock)
+        }
+
+        fun getStatistics(): OrderedJsonObject {
```
bgrozev, line 128: This could be just `= OrderedJsonObject().apply { queueStatsById.entries.forEach { put(it.key, it.value.stats) } }`. This feels easier to read, but I know that's subjective.

src/main/java/org/jitsi/utils/queue/PacketQueue.java
JonathanLennox: Yeah, I'll remove the `setObserver` case.

bgrozev (approved):

PR #102, opened 2021-06-10
Fix json debug encoding of speaker levels.

jqdrqgnq:
Encode valid json.
Encode array elements, not hashcode.

### DIFF
--- a/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
+++ b/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
@@ -1071,7 +1071,7 @@ public class DominantSpeakerIdentification<T>
          * Gets the (history of) audio levels received or measured for this
          * <tt>Speaker</tt>.
          *
-         * @return a <tt>byte</tt> array which represents the (history of) audio
+         * @return a <tt>String</tt> that lists the (history of) audio
          * levels received or measured for this <tt>Speaker</tt>
          */
         String getLevels()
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -1074,21 +1074,26 @@ public synchronized long getLastLevelChangedTime()
          * @return a <tt>byte</tt> array which represents the (history of) audio
```
JonathanLennox, line 1074: The documentation should be updated to reflect this new behavior.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -1074,21 +1074,26 @@ public synchronized long getLastLevelChangedTime()
...
          * levels received or measured for this <tt>Speaker</tt>
          */
-        byte[] getLevels()
+        String getLevels()
         {
             // The levels of Speaker are internally maintained starting with the
             // last audio level received or measured for this Speaker and ending
             // with the first audio level received or measured for this Speaker.
             // Unfortunately, the method is expected to return levels in reverse
```
JonathanLennox, line 1082: Since this is now for human consumption, I don't think this is so much "this method is expected to" as much as "we want it to" - maybe update the comment?

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -1074,21 +1074,26 @@ public synchronized long getLastLevelChangedTime()
...
-            for (int s = src.length - 1, d = 0; d < dst.length; --s, ++d)
+            for (int s = src.length - 1; s >= 0; --s)
             {
-                dst[d] = src[s];
+                sb.append(src[s]);
+                sb.append(',');
             }
-            return dst;
+
+            sb.setLength(sb.length() - 1); /* pop last comma */
```
JonathanLennox, line 1094: Could just add the comma first in the loop, with `if (s != 0) sb.append(',')`

JonathanLennox:
Jenkins, add to whitelist.

JonathanLennox (approved):

bgrozev (approved):

PR #104, opened 2021-06-25
Keep track of the loudest speakers in a conference.

jqdrqgnq:
Provide method to check if a received audio packet is from one
of the loudest speakers.

### DIFF
--- a/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
+++ b/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
@@ -549,6 +549,10 @@ public class DominantSpeakerIdentification<T>
 
         /**
          * The endpoint's current rank by energy level.
+         * If the endpoint is not in the list of the current loudest speakers,
+         * this will be set to the size of the list (which will be <tt>numLoudestToTrack</tt>
+         * if the list is at its maximum size). In essence, all untracked endpoints
+         * are considered tied for the next highest rank after the tracked ones.
          */
         public final int energyRanking;
 
--- a/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
+++ b/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
@@ -18,11 +18,11 @@ package org.jitsi.utils.dsi;
 import java.lang.ref.*;
 import java.util.*;
 import java.util.concurrent.*;
-import java.util.logging.Level;
+import java.util.logging.*;
 
 import org.jitsi.utils.concurrent.*;
 import org.jitsi.utils.logging.*;
-import org.jitsi.utils.stats.BucketStats;
+import org.jitsi.utils.stats.*;
 import org.json.simple.*;
 
 /**
@@ -334,10 +334,30 @@ public class DominantSpeakerIdentification<T>
     {
     }
 
+    /**
+      * The number of current loudest speakers to route audio packets for.
+      */
     private int numLoudestToRoute = 3;
+
+    /**
+      * Whether to route dominant speaker when it is not among the current loudest speakers.
+      */
     private boolean alwaysRouteDominant = false;
+
+    /**
+      * Time in milliseconds after which speaker is removed from loudest list if
+      * no new audio packets have been received from that speaker.
+      */
     private int energyExpireTimeMs = 150;
+
+    /**
+      * Alpha factor for exponential smoothing of energy values, multiplied by 100.
+      */
     private int energyAlphaPct = 50;
+
+    /**
+      * Distribution of energy scores for discarded audio packets
+      */
     private BucketStats tossedPacketsEnergyStats = null;
 
     public synchronized void setLoudestConfig(int numLoudestToRoute_, boolean alwaysRouteDominant_, int energyExpireTimeMs_, int energyAlphaPct_)
--- a/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
+++ b/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
@@ -528,12 +528,15 @@ public class DominantSpeakerIdentification<T>
         while (loudest.size() > numLoudestToRoute)
             loudest.remove(numLoudestToRoute);
 
-        i = 0;
-        while (i < loudest.size())
+        if (logger.isDebugEnabled())
         {
-            Speaker cur = loudest.get(i);
-            logger.debug("New list: " + i + ": " + cur.id.toString() + ": " + cur.energyScore + ".");
-            ++i;
+            i = 0;
+            while (i < loudest.size())
+            {
+                Speaker cur = loudest.get(i);
+                logger.debug("New list: " + i + ": " + cur.id.toString() + ": " + cur.energyScore + ".");
+                ++i;
+            }
         }
     }
 
--- a/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
+++ b/src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
@@ -519,7 +519,7 @@ public class DominantSpeakerIdentification<T>
             ++i;
         }
 
-        if(i < numLoudestToRoute)
+        if (i < numLoudestToRoute)
         {
             logger.debug("Adding " + speaker.id.toString() + " at position " + i + ".");
             loudest.add(i, speaker);
@@ -539,7 +539,7 @@ public class DominantSpeakerIdentification<T>
 
     public synchronized boolean isAmongLoudestSpeakers(T id)
     {
-        if(alwaysRouteDominant && dominantId != null && dominantId.equals(id))
+        if (alwaysRouteDominant && dominantId != null && dominantId.equals(id))
         {
             logger.debug("Found " + id + ". Dominant speaker.");
             return true;
@@ -560,10 +560,10 @@ public class DominantSpeakerIdentification<T>
         logger.debug(id + " not found.");
 
         BucketStats tossedPacketsEnergyStats = this.tossedPacketsEnergyStats;
-        if(tossedPacketsEnergyStats != null)
+        if (tossedPacketsEnergyStats != null)
         {
             Speaker speaker = getOrCreateSpeaker(id);
-            if(speaker != null)
+            if (speaker != null)
             {
                 tossedPacketsEnergyStats.addValue(speaker.energyScore);
             }
@@ -604,7 +604,7 @@ public class DominantSpeakerIdentification<T>
         {
             int cookedLevel = speaker.levelChanged(level, now);
 
-            if(cookedLevel >= 0)
+            if (cookedLevel >= 0)
                 updateLoudestList(speaker, cookedLevel, now);
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -319,13 +321,42 @@ private static double computeSpeechActivityScore(
...
      */
     public DominantSpeakerIdentification()
     {
     }
 
+    private int numLoudestToRoute = 3;
+    private boolean alwaysRouteDominant = false;
+    private int energyExpireTimeMs = 150;
+    private int energyAlphaPct = 50;
+    private BucketStats tossedPacketsEnergyStats = null;
```
damencho, line 341: Could you please add javadocs to all members what are those used for.

damencho (dismissed):
Please add and javadocs to the added methods, thanks.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -18,9 +18,11 @@
 import java.lang.ref.*;
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.logging.Level;
```
damencho, line 21: Change those to package imports.

jqdrqgnq: the change you suggested causes a compile error.
> both class org.jitsi.utils.logging.Logger in org.jitsi.utils.logging and class java.util.logging.Logger in java.util.logging match
any better resolution?

bbaldino: Any reason you're using `logger.log(LEVEL.INFO...`, etc. as opposed to `logger.info(` ?

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -1141,7 +1290,7 @@ public void levelChanged(int level)
          * @param time the (local <tt>System</tt>) time in milliseconds at which
          * the specified <tt>level</tt> has been received or measured
          */
-        public synchronized void levelChanged(int level, long time)
+        public synchronized int levelChanged(int level, long time)
```
JonathanLennox, line 1293: Add a doc for the new return value?

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -461,7 +602,10 @@ public void levelChanged(T id, int level)
         }
         if (speaker != null)
         {
-            speaker.levelChanged(level, now);
+            int cookedLevel = speaker.levelChanged(level, now);
+
+            if(cookedLevel >= 0)
```
JonathanLennox, line 607: Space after if

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -431,6 +482,96 @@ private synchronized Speaker getOrCreateSpeaker(T id)
         return speaker;
     }
 
+    private synchronized void updateLoudestList(Speaker speaker, int level, long now)
+    {
+        long oldestValid = now - energyExpireTimeMs;
+
+        /* exponential smoothing */
+        speaker.energyScore = (energyAlphaPct * level + (100 - energyAlphaPct) * speaker.energyScore) / 100;
```
JonathanLennox, line 490: Should the alpha factor be a declared constant?

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -431,6 +482,96 @@ private synchronized Speaker getOrCreateSpeaker(T id)
...
+
+        logger.debug("Want to add " + speaker.id.toString() + " with score " + speaker.energyScore + ". Last level = " + level + ".");
+
+        int i = 0;
+        while (i < loudest.size())
+        {
+            Speaker cur = loudest.get(i);
+            if (cur.getLastLevelChangedTime() < oldestValid)
+            {
+                logger.debug("Removing " + cur.id.toString() + ". old.");
```
JonathanLennox, line 500: I feel like these should be `trace` rather than `debug` level logs - I generally prefer trace for logs that run on every operation.

String concatenation isn't *that* expensive, but I also wonder if it's worthwhile to switch to using `Logger2` and use closures to create the log messages, so we don't do the concatenation when the logs aren't enabled.

ibauersachs: For messages as simple as these it would be really better to use a logging framework that understands template parameters, e.g. Logback or log4j2 (both via slf4j) or [Flogger](https://github.com/google/flogger).
The statement would then be `logger.debug("Removing {} old", cur.id);`, which is essentially free if debug logging is disabled (and cheaper than lambdas). And the readability gain is also massive.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
bgrozev: I think `energyAlphaPct` can't be a constant, because it's configurable?

I don't know if this is important, but integer division takes the floor of the result (so 99/100 is 0). You can achieve rounding by adding half the number being divided.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -431,6 +482,96 @@ private synchronized Speaker getOrCreateSpeaker(T id)
...
+        {
+            logger.debug("Adding " + speaker.id.toString() + " at position " + i + ".");
+            loudest.add(i, speaker);
+        }
+
+        while (loudest.size() > numLoudestToRoute)
+            loudest.remove(numLoudestToRoute);
+
+        i = 0;
+        while (i < loudest.size())
```
bgrozev, line 532: Surround this look in `if (logger.isDebugEnabled())`?

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -431,6 +482,96 @@ private synchronized Speaker getOrCreateSpeaker(T id)
...
+            {
+                logger.debug("Found " + id + " at location " + i + ".");
+                return true;
+            }
+            ++i;
+        }
+
+        logger.debug(id + " not found.");
+
+        BucketStats tossedPacketsEnergyStats = this.tossedPacketsEnergyStats;
```
bgrozev, line 562: This feels awkward. Why is the `isAmongLoudestSpeakers` function responsible for updating the dropped packet stats? I guess the reason is that you need access to `energyScore`, but is this really what you want to keep track of (as opposed to simply the level of the discaded packet)? It would be cleaner if the code dropping the packet updated the stats.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
jqdrqgnq: I wanted to log a lot for testing. I can remove them before submitting.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
jqdrqgnq: done, though these logs are probably temporary.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
jqdrqgnq: now returning a class with ranking info. should be a cleaner interface and better if there are future additions.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -431,6 +467,124 @@ private synchronized Speaker getOrCreateSpeaker(T id)
...
+            if (speaker.id.equals(id))
+            {
+                return new SpeakerRanking(isDominant, rank, speaker.energyScore);
+            }
+            ++rank;
+        }
+
+        Speaker speaker = getOrCreateSpeaker(id);
+        int energyScore = speaker != null ? speaker.energyScore : 0;
+        return new SpeakerRanking(isDominant, rank, energyScore);
```
bgrozev, line 585: This case returns `rank = loudest.size()`, even though the exact rank is unknown. Should it return `rank = -1` to make it more obvious?

jqdrqgnq: could do that. caller would need to check for the special value, though. i was also thinking this would return a valid rank in the case where the list was not at max size but the endpoint wasn't found. not sure if that could happen - maybe if all the packets from an endpoint were filtered.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
jqdrqgnq: yes it's configurable.

agreed, rounding to nearest now.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
jqdrqgnq: After further thought, I think this is okay. I interpret it to mean all untracked endpoints are tied for the next rank after the tracked ones. It avoids the caller needing to check for the special value and the danger of -1 being read as a higher rank than 0 if they don't check.
Added commentary to the docs explaining the return value in this case.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
jqdrqgnq: removed.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
jqdrqgnq: moved to logger2, used closures.

jqdrqgnq: move to trace level as well.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -319,13 +318,51 @@ private static double computeSpeechActivityScore(
...
      * Initializes a new <tt>DominantSpeakerIdentification</tt> instance.
      */
     public DominantSpeakerIdentification()
     {
     }
 
+    /**
+      * The number of current loudest speakers to keep track of.
+      */
+    private int numLoudestToTrack = 3;
```
bgrozev, line 337: Can we default to `0` to avoid the calculations unless the users need them?

jqdrqgnq: the default jvb configuration was setting this to 0, but I agree that the module should have the correct default independent of who's using it. fixed.

bgrozev (approved):

bgrozev (approved):

PR #119, opened 2022-05-02
Remove the guava dependency

bgrozev:
- ref: Move LogContext to kt (remove guava dep).
- ref: Port LogContextTest to kt, deprecate JMap.


src/main/java/org/jitsi/utils/collections/JMap.java
```diff
@@ -27,6 +27,7 @@
  * taking individual arguments--are in Java 9 and are provided here
  * for convenience.
  */
+@Deprecated
```
JonathanLennox, line 30: Add a comment saying that since we now target Java 11+, this should be avoided in favor of using the native Java methods?

src/main/kotlin/org/jitsi/utils/logging2/LogContext.kt
```diff
@@ -0,0 +1,97 @@
...
+        set(newValue) {
+            field = newValue
+            updateFormattedContext()
+        }
+
+    /**
+     * The formatted String representing the total context (the combination of the ancestors' context and this
+     * context)
+     */
+    var formattedContext: String = formatContext(context)
```
JonathanLennox, line 46: Isn't this initialization redundant with the `init` below?

src/main/java/org/jitsi/utils/collections/JMap.java
bgrozev: I'll just remove it now and update the projects.

JonathanLennox: It would mean that jicoco and ice4j need to be updated to Java 11, but if you're good with that, that's fine.

JonathanLennox (approved):

src/main/java/org/jitsi/utils/collections/JMap.java
bgrozev: No, I don't want to force them to 11 yet, so I used a different API there.

PR #120, opened 2022-05-20
ref: Change the JSON format for BucketStats, make toJson open

bgrozev:


bgrozev:
This avoids special characters in the JSON field names: ' ', '<', '>', '='. The new format looks like

```
   "conference_sizes": {
        "average": null,
        "max": 0,
        "total_value": 0,
        "total_count": 0,
        "buckets": {
            "0": 0,
            "1": 0,
            "2": 0,
            "3_to_5": 0,
            "6_to_10": 0,
            "11_to_20": 0,
            "21_to_50": 0,
            "51_to_100": 0,
            "101_to_200": 0,
            "201_to_300": 0,
            "301_to_400": 0,
            "401_to_500": 0,
            "501_to_max": 0,
            "p99_upper_bound": -1,
            "p999_upper_bound": -1
        }
    },
```

JonathanLennox (dismissed):

JonathanLennox:
We'll have to remember to chase the various consumers of this.

A question - I think most cases where we're using this, we actually compute cumulative values.  Would it be better to change this to report cumulative values?

src/main/kotlin/org/jitsi/utils/stats/BucketStats.kt
```diff
@@ -33,31 +36,33 @@ open class BucketStats(thresholdsNoMax: LongArray, val averageMaxLabel: String =
...
+        totalCount.increment()
+        buckets.addValue(value)
     }
 
-    open fun toJson() = OrderedJsonObject().apply {
+    open fun toJson(format: Format = Format.Separate) = OrderedJsonObject().apply {
         val snapshot = snapshot
         put("average$averageMaxLabel", snapshot.average)
         put("max$averageMaxLabel", snapshot.maxValue)
+        put("min$averageMaxLabel", snapshot.minValue)
```
JonathanLennox, line 54: Maybe name this `averageMaxMinLabel` now?

JonathanLennox (commented):
Note that the commit description is wrong - `toJson` was always open, what you changed was to make `getBucketsJson` open.  (Why, btw?)

src/test/kotlin/org/jitsi/utils/stats/BucketStatsTest.kt
```diff
@@ -50,21 +62,47 @@ class BucketStatsTest : ShouldSpec() {
...
-                bucketsJson["4_to_5_ms"] shouldBe 100
-                bucketsJson["6_to_200_ms"] shouldBe 1
-                bucketsJson["201_to_999_ms"] shouldBe 0
-                bucketsJson["1000_to_max_ms"] shouldBe 1
+
+                context("In separate buckets (default)") {
+                    val bucketsJson = bucketStats.toJson()["buckets"]
+                    bucketsJson.shouldBeInstanceOf<OrderedJsonObject>()
+
+                    bucketsJson["min_to_0_ms"] shouldBe 1
```
JonathanLennox, line 76: For items that are inherently unsigned (like number of participants in a conference, having "min_to_0" as the first bucket is a bit weird?

bgrozev: I changed it to not automatically add Long.MIN_VALUE and Long.MAX_VALUE to the intervals, so we can start at 0 in such cases.

src/main/kotlin/org/jitsi/utils/stats/BucketStats.kt
```diff
@@ -184,4 +206,4 @@ class Buckets(thresholdsNoMax: LongArray) {
 }
 
 /** Utility class with thresholds suitable for recording conference sizes. */
-class ConferenceSizeBuckets : BucketStats(longArrayOf(0, 1, 2, 5, 10, 20, 50, 100, 200, 300, 400, 500))
+class ConferenceSizeBuckets : BucketStats(listOf(0, 1, 2, 5, 10, 20, 50, 100, 200, 300, 400, 500))
```
JonathanLennox: If you're removing the automatic MAX do you want to add it here?

bgrozev: I did, in a later commit

JonathanLennox (approved):
Don't forget to fix the commit message (getBucketsJson was made open, not toJson).

PR #123, opened 2022-07-18
Refactor DSI and add tests

bgrozev:
This changes `DominantSpeakerIdentification` to use a `ScheduledExecutorService` instead of `Thread.sleep` for scheduling. Then pulls in `FakeClock` and `FakeScheduledExecutorService` from jicoco in order orchestrate a test for DSI (added a trace file from a test conference).


src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -921,54 +920,42 @@ public DecisionMaker(DominantSpeakerIdentification algorithm)
...
+                // commit suicide.
+                if (sleep < 0)
+                {
+                    exit = true;
+                }
+                else
+                {
+                    // Before sleeping, make the currentThread release
+                    // its reference to the associated
+                    // DominantSpeakerIdentification instance.
```
JonathanLennox, line 947: I don't think this comment still applies?  It looks like it referred to the `algorithm = null` in the old code, which isn't relevant since the task isn't persisting.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -921,54 +920,42 @@ public DecisionMaker(DominantSpeakerIdentification algorithm)
...
+                    algorithm.executor.schedule(this, sleep, TimeUnit.MILLISECONDS);
                 }
-                while (true);
             }
-            finally
+
+            if (exit)
             {
                 // Notify the algorithm that this background thread will no
```
JonathanLennox, line 954: Comment shouldn't refer to a background thread anymore.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -921,54 +920,42 @@ public DecisionMaker(DominantSpeakerIdentification algorithm)
...
                 // longer run it in order to make the (global) decision about
                 // speaker switches. Subsequently, the algorithm may decide to
                 // spawn another background thread to run the same task.
```
JonathanLennox, line 957: This comment shouldn't refer to a background thread anymore.

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -921,54 +920,42 @@ public DecisionMaker(DominantSpeakerIdentification algorithm)
...
+                    // Before sleeping, make the currentThread release
+                    // its reference to the associated
+                    // DominantSpeakerIdentification instance.
+                    algorithm.executor.schedule(this, sleep, TimeUnit.MILLISECONDS);
                 }
-                while (true);
             }
-            finally
+
+            if (exit)
```
JonathanLennox, line 952: The old code would call `algorithm.decisionMakerExited()` if an exception was thrown - you may still need that?

JonathanLennox (commented):
There's a spotbugs failure in FakeScheduledExecutorService - probably something we're suppressing in jicoco?

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -932,29 +932,21 @@ public void run()
...
                 }
             }
 
             if (exit)
             {
-                // Notify the algorithm that this background thread will no
-                // longer run it in order to make the (global) decision about
-                // speaker switches. Subsequently, the algorithm may decide to
-                // spawn another background thread to run the same task.
+                // Notify the algorithm that this DecisionMaker no longer run. Subsequently, the algorithm may decide
```
JonathanLennox, line 957: Syntax, "will no longer run"

src/main/kotlin/org/jitsi/utils/concurrent/FakeScheduledExecutorService.kt
```diff
@@ -0,0 +1,232 @@
...
+                    run()
+                } else {
+                    // The job wasn't ready yet, re-add it to the front of the queue
+                    jobs.add(0, job)
+                }
+            }
+        }
+    }
+
+    /* Methods required by ScheduledExecutorService, but not needed by our unit tests. */
```
JonathanLennox, line 133: This comment should be right before the block of TODOs

JonathanLennox (approved):

JonathanLennox (approved):

PR #126, opened 2022-07-21
Add eldest() to LRUCache class

jqdrqgnq:


JonathanLennox (approved):

src/main/java/org/jitsi/utils/LRUCache.java
```diff
@@ -68,6 +68,11 @@ public LRUCache(int cacheSize, boolean accessOrder)
         this.cacheSize = cacheSize;
     }
 
+    /**
+     * @return the eldest item in the cache.
+     */
+    public Map.Entry<K, V> eldest() { return this.entrySet().iterator().next(); }
```
bgrozev, line 74: How does this work with `accessOrder=true` (i.e. does it always return the item that is the next one to be replaced?)? Maybe add some tests in LRUCacheTest.

jqdrqgnq: yes, in both modes it's the next to be replaced. do you find the method name confusing? same terminology is used by another method, though granted that one isn't public. i can add tests.

bgrozev: > do you find the method name confusing?

No, not at all.

bgrozev (approved):

PR #137, opened 2024-05-08
Use logRecord.getInstant() for timestamp in JitsiLogFormatter logs.

JonathanLennox:
Also use DateTimeFormatter to format the timestamp, and some other warning cleanups.


JonathanLennox:
LogRecord.getInstant() was added in Java 9, so we can use it now that we're on 11+.

damencho (approved):

PR #139, opened 2024-09-05
Add some more Duration and Instant helper utils

JonathanLennox:
Some of these are moved from jitsi-media-transform's utils directory.

### DIFF
--- a/src/main/kotlin/org/jitsi/utils/Duration.kt
+++ b/src/main/kotlin/org/jitsi/utils/Duration.kt
@@ -17,13 +17,17 @@
 package org.jitsi.utils
 
 import java.time.Duration
+import java.time.temporal.ChronoUnit
 
 /**
- * Helpers to create instance of [Duration] more easily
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/jitsi/utils/Duration.kt
```diff
@@ -40,5 +44,59 @@ val Number.mins: Duration
 val Number.days: Duration
     get() = Duration.ofDays(this.toLong())
 
-operator fun Duration.times(x: Int): Duration = Duration.ofNanos(toNanos() * x)
+operator fun Duration.times(x: Number): Duration = this.multipliedBy(x.toLong())
```
bgrozev, line 47: This accepts a `Number` but will not work as expected with e.g. `1.5`. I see that `kotlin.time.Duration` already has a `times(scale: Double)`. Can we use kotlin Duration everywhere?

bgrozev (dismissed):

bgrozev (approved):

PR #146, opened 2025-06-11
feat: Add a URL template utility.

bgrozev:


src/test/kotlin/org/jitsi/utils/TemplatedUrlTest.kt
```diff
@@ -0,0 +1,95 @@
...
+
+        should("resolve a template with multiple keys when they are not set()") {
+            val template = "wss://{{host}}/{{path}}?param={{param}}"
+            val templatedUrl = TemplatedUrl(template)
+            templatedUrl.set("host", "example.com")
+            templatedUrl.set("path", "api/resource")
+            templatedUrl.set("param", "value")
+
+            templatedUrl.resolve(
+                mapOf("host" to "example.com", "path" to "api/resource", "param" to "value")
```
JonathanLennox, line 52: Do you want these to have different values to ensure that it's picking up the right ones?

src/test/kotlin/org/jitsi/utils/TemplatedUrlTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        should("throw an exception when not all requiredKeys are set") {
+            val template = "https://example.com/{{path}}?param={{param}}"
+            val templatedUrl = TemplatedUrl(template, requiredKeys = setOf("param"))
+            templatedUrl.set("path", "api")
+
+            shouldThrow<IllegalArgumentException> {
+                templatedUrl.resolve()
+            }
+        }
+        should("throw an exception when the URI is invalid") {
```
JonathanLennox, line 78: It'd probably also be good to test that we throw an exception if the values are present but generate an invalid url.

src/test/kotlin/org/jitsi/utils/TemplatedUrlTest.kt
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+
+        should("handle keys that appear multiple times") {
+            val template = "https://{{host}}/{{path}}/{{path}}"
+            val templatedUrl = TemplatedUrl(template)
+            templatedUrl.set("host", "example.com")
+            templatedUrl.set("path", "resource")
+
+            templatedUrl.resolve() shouldBe URI("https://example.com/resource/resource")
```
JonathanLennox, line 92: You don't have a test for the `requiredKeys` parameter

src/main/kotlin/org/jitsi/utils/TemplatedUrl.kt
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.utils
+
+import java.net.URI
+
+class TemplatedUrl(
+    private val template: String,
+    private val map: MutableMap<String, String> = mutableMapOf(),
```
JonathanLennox, line 22: If someone passes in a MutableMap they can then manipulate it outside of this object, which feels like a bad idea.  Make the argument a map, and copy it to a private mutableMap?

src/test/kotlin/org/jitsi/utils/TemplatedUrlTest.kt
```diff
@@ -0,0 +1,95 @@
...
+            val template = "https://example.com/{{path}}?param={{param}}"
+            val templatedUrl = TemplatedUrl(template, requiredKeys = setOf("param"))
+            templatedUrl.set("path", "api")
+
+            shouldThrow<IllegalArgumentException> {
+                templatedUrl.resolve()
+            }
+        }
+        should("throw an exception when the URI is invalid") {
+            val templatedUrl = TemplatedUrl("https://example.com/{{}}")
```
JonathanLennox, line 79: I take it this is relying on the fact than an unquoted `{` or `}` isn't valid anywhere in a URI?

JonathanLennox (approved):

PR #158, opened 2026-06-01
Replace json-simple with Jackson

bgrozev:
Replace the `org.json.simple` dependency with Jackson (`jackson-databind`).

- `OrderedJsonObject` is now a factory function returning `ObjectNode` instead of a class
- Update `jackson-databind` to 2.19.0
- Add SpotBugs exclusion for the factory function naming convention


src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -432,50 +435,45 @@ synchronized void decisionMakerExited(DecisionMaker decisionMaker)
...
      * </p>
      *
-     * @return a <tt>JSONObject</tt> which represents this instance of the
-     * purposes of the REST API of Videobridge
+     * @return a {@link JsonNode} which represents this instance for the
+     * purposes of the REST API of Videobridge, or {@code null} if DEBUG is off.
      */
-    @SuppressWarnings("unchecked")
-    public JSONObject doGetJSON()
+    public JsonNode doGetJSON()
```
JonathanLennox, line 442: Annotate `@Nullable`?

src/main/kotlin/org/jitsi/utils/queue/QueueStatistics.kt
```diff
@@ -28,6 +29,8 @@ import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.LongAdder
 import kotlin.collections.ArrayList
 
+private val jsonMapper = ObjectMapper()
```
JonathanLennox: I think for Kotlin the `jacksonObjectMapper()` helper function is preferred, though I'm not sure if it's relevant for your use cases here.

bgrozev: jacksonObjectMapper comes from jackson-module-kotlin which is not currently a dependency of jitsi-utils

src/main/java/org/jitsi/utils/dsi/DominantSpeakerIdentification.java
```diff
@@ -424,6 +425,8 @@ synchronized void decisionMakerExited(DecisionMaker decisionMaker)
         }
     }
 
+    private static final ObjectMapper MAPPER = new ObjectMapper();
```
JonathanLennox, line 428: You're still creating a MAPPER here.

src/main/java/org/jitsi/utils/logging/TimeSeriesLogger.java
```diff
@@ -26,6 +26,8 @@
  */
 public class TimeSeriesLogger
 {
+    private static final ObjectMapper MAPPER = new ObjectMapper();
```
JonathanLennox, line 29: Whereas this one I think you need an actual `ObjectMapper` because you're using `valueToTree`?

spotbugs-exclude.xml
```diff
@@ -46,4 +46,5 @@
     <Source name="~.*\.*Test.java" />
   </Or>
 </Match>
+
```
JonathanLennox: You've now added some extra whitespace here

JonathanLennox (approved):

